PR #72, opened 2023-07-06
switching our facebook example to use hf modules and basaran

brandonjbjelland:


samos123:
I have a large PR out that makes this change already: https://github.com/substratusai/substratus/pull/49
could we merge that instead once I'm done with it? I rather not have to resolve conflicts if possible.

nstogner (approved):

samos123:
this can be closed since PR #49 that does this was already merged

PR #73, opened 2023-07-06
WIP: attempting to port forward 8889 also. hangs on waiting

brandonjbjelland:


.gitignore
```diff
@@ -59,6 +59,5 @@ override.tf.json
 .terraformrc
 terraform.rc
 
-open-notebook
```
samos123, line 62: what's the reason for removing this from .gitignore? I would assume it should stay ignored?

kubectl/open-notebook/client.go
```diff
@@ -107,11 +110,45 @@ func (c *notebookClient) portForward(ctx context.Context, verbose bool, nb *unst
...
 		stdout, stderr = io.Discard, io.Discard
 	}
-	fw, err := portforward.New(dialer, []string{fmt.Sprintf("%d:%d", localPort, podPort)}, ctx.Done(), ready, stdout, stderr)
-	if err != nil {
-		return err
+
+	// TODO: Remove hardcoding.
+	portPairs := []portPair{
+		{containerPort: 8888, localPort: 8888},
+		{containerPort: 8889, localPort: 8889},
```
samos123, line 117: shouldn't the container port stay 8888?

samos123 (changes requested):
looks good overall, had some question that might require changes.

.gitignore
brandonjbjelland: we have a directory at `kubectl/open-notebook` and git git got unhappy when I changed files in that path and added them. Now that our docs are explicit about moving the binary, I think removing this is okay.

An alternative: can we name either of the directory or binary differently?

kubectl/open-notebook/client.go
brandonjbjelland: If we want to support serving both jupyter lab via the browser (8888) and a jupyter notebook server for an IDE to plug in, we need to forward another port (8889).

It turns out this pattern isn't vscode specific. [pycharm](https://www.jetbrains.com/help/pycharm/configuring-jupyter-notebook.html) also supports the same pattern.

brandonjbjelland:
Heh! Yet another PR I'm going to close outright here. Nick and I got this working but in the end we decided to NOT run and expose 2 ports and 2 services but instead just run `jupyter notebook` from the container. As an option, when creating the notebook object we could add a parameter to specify you want the lab interface instead of server.

PR #75, opened 2023-07-07
Change the world refactor

### LINKED ISSUE #68: Store models outside of containers
Storing models outside of containers could help with:

1. Loading models into different environments. (A notebook/dev environment could be separate from a production serving environment).
2. Allowing for users to iterate on the same models with different libraries (see #67).

A Model object in this case would operate closer to a Dataset (where an importer is defined that will load in a model to be saved).

Underlying storage would likely be a FUSE-mounted bucket.

nstogner:
* New API Design
* Move Model storage outside of container (Fixes #68)
* Remove parameter based GPU memory calculations
* Refactor tests
* Add container reconciler
* etc
* etc
* etc


docs/design.md
```diff
@@ -0,0 +1,67 @@
...
+### Model Sources
+
+A Model can be sourced from Git. This will trigger a build Job.
+
+```yaml
+kind: Model
+spec:
+  source:
+    git:
+      url: https://github.com/substratusai/model-falcon-7b
```
samos123, line 27: I guess in this case it could be 
```
url: https://github.com/substratusai/model-loader-huggingface
name: falcon-7b
```

The huggingface model loader can load any model from huggingface into the GCS bucket

docs/design.md
```diff
@@ -0,0 +1,67 @@
...
+A Model object represents a logical ML model.
+
+The object describes where the source code for the model should come from. This code will be stored in a container image. That container image will initially be run in order to store the initial weights and biases in a bucket that is mounted into the container as a directory. This is most useful in cases where the Model is representing a pre-trained model.
+
+The Model object can optionally describe the training parameters including a reference to the Dataset that should be used for training. In this case, the Model object represents the eventual output of a training/finetuning operation.
+
+```yaml
+kind: Model
+spec:
+  source: {} # Required
```
samos123, line 14: could we make source optional such that you can create a notebook with an empty model attached?

samos123: That would also allow us to have a FinetuningJob CRD later if needed

docs/design.md
```diff
@@ -0,0 +1,153 @@
...
+
+  trainer:
+    # sourceModel can be left out for creating new base model
+    sourceModel:
+      name: falcon-7b
+      namespace: other-namespace
+    dataset:
+      name: k8s-instructions
+      namespace: other-namespace
+    resources: {}
```
samos123, line 80: should resources be under `container`?

docs/design.md
```diff
@@ -0,0 +1,153 @@
...
+3. Experiment with the model and training and only save my notebook somewhere for reference
+
+```yaml
+kind: Notebook
+name: notebooks-for-anything
+spec:
+  container:
+    # or git repo as a source
+    image: substratusai/notebook-gpu
+  resources: {gpuCount: 4, type: L4}
```
samos123, line 107: I think resources should be exactly the same as container resources:
```
    resources:
      requests:
        memory: "10Gi"
        cpu: "10"
      limits:
       nvidia.com/gpu: 2
```
would be valid here.

The open question then would be how to pass it GPU type. Should that be a separate attribute outside of resources?  I vote for yes gpu_type has separate attribute such that we can use standard K8s resource requests which most are familiar with.

api/v1/model_types.go
```diff
@@ -6,47 +6,42 @@ import (
...
+	// Trainer should be set to run a training job.
+	Trainer *ModelTrainer `json:"trainer,omitempty"`
 }
 
-// +kubebuilder:validation:Enum=CPU;GPU
-type ComputeType string
+func (m *Model) GetContainer() Container {
+	return m.Spec.Container
+}
+func (m *Model) SetContainer(c Container) {
```
samos123, line 22: it doesn't look like SetContainer gets used anywhere?

internal/builder/builder.go
```diff
@@ -0,0 +1,279 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	saName = "container-builder"
+)
+
+type BuildableObject interface {
+	client.Object
+	GetContainer() apiv1.Container
```
samos123, line 30: should GetContainer return a pointer such that you don't need another SetContainer?

internal/builder/builder.go
```diff
@@ -0,0 +1,279 @@
...
+		})
+		if err := r.Client.Status().Update(ctx, obj); err != nil {
+			return Result{}, fmt.Errorf("updating status: %w", err)
+		}
+
+		// Allow Job watch to requeue.
+		return Result{}, nil
+	}
+
+	container := obj.GetContainer()
```
samos123, line 107: GetContainer returns a copy of the Container object so the next few lines shouldn't be able to update the actual Image attribute of the underlying `obj`?

docs/design.md
```diff
@@ -0,0 +1,153 @@
...
+```yaml
+kind: Model
+name: falcon-7b
+spec:
+  container:
+    git:
+      url: https://github.com/substratusai/hf-model-loader
+    image: substratusai/hf-model-loader
+  loader:
+    params: {name: tiuu/falcon-7b}
```
samos123, line 39: can we convert any param.x to PARAM_X as env variable?

So params.name would be available as env variable `PARAM_NAME` in the container itself.

What you think? @brandonjbjelland @nstogner

api/v1/model_types.go
```diff
@@ -6,47 +6,40 @@ import (
...
-type ModelTraining struct {
-	// DatasetName is the .metadata.name of the Dataset to use for training.
-	DatasetName string `json:"datasetName"`
+type ModelTrainer struct {
+	// Dataset to mount for training.
+	Dataset ObjectRef `json:"datasetName"`
+
 	// Params is a list of hyperparameters to use for training.
-	Params ModelTrainingParams `json:"params"`
+	Params ModelTrainerParams `json:"params"`
```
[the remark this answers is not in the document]
nstogner: I thought about this... What about well-known params like epochs? I wonder if it makes sense to have these strongly-typed and a part of the API schema? This would encourage the use of the standard params, which would help with things like automated hyperparameter tuning down the road.

docs/design.md
nstogner: I am not sure we should directly expose kubernetes `.resources`. Mainly because of what you called out here... the use of specialized equipment like GPUs requires a slew of other fields to be set. I think substratus should simplify this by exposing a few high-level resource knobs. Perhaps:

```yaml
runtime:
  compute: ["GPU", "CPU"] # Can be run on either GPUs or CPUs.
  strategy: Fastest # or "Cheapest", or "Balanced"
  memory: "10Gi" # Translates to AT LEAST 10Gi of gpu-memory and 10Gi or cpu-memory. If more is provisioned (commonly the case), it will be reported to the container using environment variables.
  disk: "50Gi"
```

CPU cores are not really a requirement... more will be faster.

api/v1/model_types.go
nstogner: Removed

internal/builder/builder.go
nstogner: Updated it to do this

docs/design.md
nstogner: https://github.com/substratusai/substratus/pull/75#discussion_r1257490669

api/v1/model_types.go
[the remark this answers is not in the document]
samos123: What if some end-user needs to pass more params to their training? My take is we should move any strongly typed well known params 1 level up and leave Params to be dynamic in nature.

docs/design.md
```diff
@@ -0,0 +1,153 @@
...
+    # optional future
+    paramsFrom: configMap?
```
samos123, line 41:
```suggestion
  status:
    url: gcs://my-bucket/my-model/
    files: ["pytorch-001.bin", "config.json"]
    size: 27Gi
```

docs/design.md
samos123: When you define a model you have to decide if you want to run the loader on CPU or GPU

Alternative proposal for runtime under loader or trainer section
```
runtime:
  accelerator: # optional if not specified will only use CPU
     type: nvidia-tesla-a100
     count: 8
  cpu: 2 # default will always be 2 if not specified
  memory: "10Gi" # will always be 10 if not specified. Translates to AT LEAST 10Gi of gpu-memory and 10Gi or cpu-memory. If more is provisioned (commonly the case), it will be reported to the container using environment variables.
  disk: "50Gi" # will always be 50Gi if not specified
```

api/v1/model_types.go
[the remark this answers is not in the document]
nstogner: I think thats a good call, either that or have `.params` and `.customParams`.

docs/design.md
nstogner: When the specify the accelerator in the Model object, the Model object is no longer portable across clusters that have access to different accelerators.

samos123: I want to specify exactly the resources for my training job. The model loaded in GCS is still portable across clusters. I think I'm misunderstanding the portability concern

brandonjbjelland: Perhaps there's a middleground where `runtimeProfile`s can be created by administrators and used by consumers in the form you're pitching @nstogner . Personally, I lean toward starting with the more raw resource spec model since model requirements are quite specific.

api/v1/model_types.go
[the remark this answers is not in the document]
samos123: I prefer to error on flexibility here instead of requiring further API changes down the line.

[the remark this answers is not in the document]
brandonjbjelland: I think we should start flexible and then back out the most common hyperparameters into strongly typed fields as they emerge. Maybe something like:

```suggestion
type ModelTrainer struct {
	// Dataset to mount for training.
	Dataset ObjectRef `json:"datasetName"`

	// Params is a list of hyperparameters to use for training.
	Params ModelTrainerParams `json:"params"`
}

type ModelTrainerParams struct {
	//+kubebuilder:default:=3
	// Epochs is the total number of iterations that should be run through the training data.
	// Increasing this number will increase training time.
	BatchSize int64 `json:"batchSize,omitempty"`
	Custom map[string]interface{} `json:"custom,omitempty"`
}
```

docs/design.md
```diff
@@ -0,0 +1,153 @@
...
+      name: falcon-7b
+      namespace: other-namespace
+    dataset:
+      name: k8s-instructions
+      namespace: other-namespace
+    resources: {}
+    params: {epochs: 1}
+```
+
+
```
brandonjbjelland, line 84:
```suggestion


```

docs/design.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Model Training with Notebook
+Notebook container contract is to run notebook.sh and serve
+notebook compatible HTTP endpoint on port 8888.
+
+User flow: Random experiments
+
+notebook-gpu image has transformers, pytorch, cuda, python 3, etc.. basically most you need
```
brandonjbjelland, line 92:
```suggestion
notebook-gpu image has transformers, pytorch, cuda, python 3, etc.. most everything you need to get started.
```

api/v1/model_types.go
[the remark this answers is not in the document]
samos123: my preference is to not have any strongly types params to improve flexibility and support all kinds of training jobs. Any param should be optional and have a default.

api/v1/common_types.go
```diff
@@ -0,0 +1,25 @@
...
+	// Git is a reference to a git repository containing model code.
+	Git *GitSource `json:"git,omitempty"`
+	// Image of a container that can be used to run the model.
+	Image string `json:"image,omitempty"`
+}
+
+type GitSource struct {
+	// URL to the git repository.
+	// Example: https://github.com/my-username/my-repo
+	URL string `json:"url,omitempty"`
```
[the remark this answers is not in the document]
nstogner: Good catch

internal/controller/utils.go
```diff
@@ -34,3 +51,146 @@ const (
 func roundUpGB(bytes int64) int64 {
 	return int64(math.Ceil(float64(bytes)/float64(gigabyte))) * gigabyte
 }
+
+type Object interface {
```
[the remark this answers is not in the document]
nstogner: Object was old, removing

docs/design.md
brandonjbjelland: > I want to specify exactly the resources for my training job. The model loaded in GCS is still portable across clusters. I think I'm misunderstanding the portability concern

Guessing at what Nick meant here: where might models not achieve portability:
1. `resource_limits` might differ from cluster to cluster.
2. different accelerator types are supported in different regions/zones and across different cloud providers.
3. even in the case where both of the above criteria are met, stock out might mean another accelerator type must be used if the job is to proceed.

I think at the moment I still lean toward raw resource specs as @samos123 is pushing for here but potentially a way to think about this is giving an ordered list of hardware specs so if one resource profile fails, another one can be tried.

docs/design.md
```diff
@@ -0,0 +1,153 @@
...
+name: notebooks-for-anything
+spec:
+  container:
+    # or git repo as a source
+    image: substratusai/notebook-gpu
+  resources: {gpuCount: 4, type: L4}
+```
+
+
+User flow: Iterate on model training code
```
brandonjbjelland, line 114: I think this captures the use case a bit more precisely:

```suggestion
Lab user flow: Iterate on model training code using a customized trainer image
```

docs/design.md
```diff
@@ -0,0 +1,153 @@
...
+
+1. `git clone https://substratusai/hf-llm-trainer && cd hf-llm-trainer`
+2. Modify Dockerfile
+2. Create a Notebook that mounts source model
+3. `kubectl open notebook notebook-training-experiment`
+   * tar up directory respecting dockerignore
+   * Get GCS signed url from controller notebook status
+   * upload to workspace.tar to signed URL
+3. Notebook automatically has all my local directory in the notebook when using the kubectl plugin
```
brandonjbjelland, line 123: This probably looks different in the VSCode extension-based flow but using `jupyter lab`:

```suggestion
1. `git clone https://substratusai/hf-llm-trainer && cd hf-llm-trainer`
2. Modify Dockerfile
3. Create a Notebook manifest and apply the Notebook CR mounting a source model
4. `kubectl open notebook -f notebook-training-experiment` - the `kubectl` plugin does each of the following additional steps when `spec.container.local = true`
   * tar up directory respecting dockerignore
   * Get GCS signed url from controller notebook status 
   * upload to workspace.tar to signed URL
5. Notebook automatically has the local directory contents in the notebook when using the `kubectl` plugin (via cp sync operations)
```

docs/design.md
```diff
@@ -0,0 +1,153 @@
...
+spec:
+  container:
+    image: substratusai/basaran # or git repo
+  # model gets mounted to /model/saved
+  model: falcon-7b-k8s
+  # optional
+  quantization:
+    bits: 4 # default is 16
+  resources: {}
+```
```
brandonjbjelland, line 156: I like this in particular! 👏

docs/design.md
nstogner: Currently everything under container is a source of the container... this would break that pattern. I vote `.spec.resources` because that way we can keep it consistent across all object kinds.

docs/design.md
```diff
@@ -0,0 +1,153 @@
...
+  resources: {}
+status:
+    uploadUrl: gs://substratusai-notebooks/workspace.tar
+```
+
+
+How to get from Notebook to trainer?
+
+Sync back files from the notebook into local directory
+and built into new image.
```
brandonjbjelland, line 141: This is an unfinished thought as we somewhat ran out of steam but here's my take:

```suggestion
User flow: How to get from Notebook to trainer?

1. As `kubectl open notebook ...` is terminated by the user, files will be synced from the Notebook back to the local directory. We can hope but not guarantee this is a git repo.
2. Optional: During `kubectl open notebook ...` termination, a signal will be sent to the controller to build a container image which can serve as the trainer.
3. Potentially the user is prompted for this auto-generation: A `model-${epoch}.yaml` manifest having the values of that trainer will be generated into the repo root having either spec.container.image populated with the build from step 2 OR the repo/branch info taken from the local dir. 

  ```yaml
  kind: Model
  name: falcon-7b-k8s-${epoch_or_branch_name}
  spec:
    container:
      image: my-org/model-falcon-7b-k8s-trainer # the just-built trainer image
  
    trainer:
      sourceModel:
        name: falcon-7b
        namespace: other-namespace
      dataset:
        name: k8s-instructions
        namespace: other-namespace
      resources: {}
      params:
        epochs: 1
  ```

4. The user could be provided a command to submit their training job.

File: internal/controller/model_controller.go [hunk 19]
```diff
@@ -35,14 +36,15 @@ const (
...
 	modelTrainerServiceAccountName = "model-trainer"
-	modelBuilderServiceAccountName = "model-builder"
 )
 
 // ModelReconciler reconciles a Model object.
 type ModelReconciler struct {
 	client.Client
 	Scheme *runtime.Scheme
 
+	*builder.ContainerReconciler
```
brandonjbjelland: This covers both use cases of trainer and loader, right?

internal/controller/runtime_test.go
```diff
@@ -345,3 +346,4 @@ func testSpec(t *testing.T, runtime Runtime) *corev1.PodSpec {
 		return nil
 	}
 }
+*/
```
brandonjbjelland, line 349: TODO: fix these or create some new tests, ya?

brandonjbjelland:
Github lost track of the files I checked as viewed... I'll consider my review pass complete and come back when you're wrapped up. Overall it LGTM

api/v1/model_types.go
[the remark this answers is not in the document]
nstogner: Agreed on defaults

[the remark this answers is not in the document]
nstogner: I like the proposal... I currently have `.custom` implemented as `.params` and moved the standard ones (`.epoch`) up to the `.trainer` level... I like having all hyperparams organized under `.params` though.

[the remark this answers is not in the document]
nstogner: @samos123 If we dont standardize some common hyperparameters as typed values, I worry that we will miss out on hyperparameter tuning functionality in the future.

samos123:
Trying out this latest branch but hitting an issue:
```
k apply -f examples/falcon-7b-instruct/base-model.yaml 
The Model "falcon-7b-instruct" is invalid: status.ready: Required value
```

This is my base-model.yaml:
```
apiVersion: substratus.ai/v1
kind: Model
metadata:
  name: falcon-7b-instruct
spec:
  container:
    git:
      url: https://github.com/substratusai/model-loader-huggingface
  loader:
    params:
      name: tiiuae/falcon-7b-instruct
```

api/v1/model_types.go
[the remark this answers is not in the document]
brandonjbjelland: > @samos123 If we dont standardize some common hyperparameters as typed values, I worry that we will miss out on hyperparameter tuning functionality in the future.

If we start untyped, we're still have a non-breaking API change away from adding typed hyperparams. I realize we're very pre-1.0 but the opposite direction would be API breaking.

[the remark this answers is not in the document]
samos123: You want hyperparameters? then use our standard naming scheme otherwise miss out on a feature.  I just don't like enforcing parameters in the API if that limits potential future use cases. I think it's too early to start enforcing parameters. Any parameter (including epochs)  should be optional and can be ignored by a trainer container.

api/v1/common_types.go
```diff
@@ -0,0 +1,55 @@
...
+	// Name of Kubernetes object.
+	Name string `json:"name"`
+
+	// FUTURE: Possibly allow for cross-namespace references.
+	// FUTURE: Possibly allow for cross-cluster references.
+}
+
+type Resources struct {
+	// CPU resources.
+	CPU *CPUResources `json:"cpu,omitempty"`
```
samos123, line 31: alternative:
```
Resources:
  CPU
  Memory:
  GPU:
    count:
     type:
```

api/v1/notebook_types.go
```diff
@@ -6,25 +6,53 @@ import (
...
+	return n.Status.Ready
+}
+func (n *Notebook) SetStatusReady(r bool) {
+	n.Status.Ready = r
 }
 
 // NotebookStatus defines the observed state of Notebook
 type NotebookStatus struct {
+	// Ready indicates that the Notebook is ready to serve. See Conditions for more details.
+	Ready bool `json:"ready"`
```
samos123, line 46: needs to add omitempty otherwise it becomes required

api/v1/modelserver_types.go
```diff
@@ -6,21 +6,29 @@ import (
...
+	Resources *Resources `json:"resources,omitempty"`
+
+	// Model references the Model object to be served.
+	Model ObjectRef `json:"model,omitempty"`
 }
 
 // ModelServerStatus defines the observed state of ModelServer
 type ModelServerStatus struct {
+	// Ready indicates whether the ModelServer is ready to serve traffic. See Conditions for more details.
+	Ready bool `json:"ready"`
```
samos123, line 22: add omitempty

brandonjbjelland: ditto: https://github.com/substratusai/substratus/pull/75#discussion_r1259334354

internal/controller/model_controller.go
nstogner: Yep, it just builds images with kaniko

internal/controller/runtime_test.go
nstogner: runtime.go got replaced by resources pkg... We should recreate these in there eventually. Not attempting in this PR.

docs/design.md
nstogner: Resolved in slack

brandonjbjelland:
Resolved the small merge conflict with main to get tests running. If we're all green, I think we're good to merge this.

brandonjbjelland:
You're a gd hero, @nstogner ! 🦸

PR #77, opened 2023-07-07
THROWAWAY model refactoring

samos123:


api/v1/common.go
```diff
@@ -0,0 +1,14 @@
...
+// TODO add validation to ensure only git, image or local is specified
+type Container struct {
+	Git   *GitSource `json:"git,omitempty"`
+	Image string     `json:"image,omitempty"`
+}
+
+type GitSource struct {
+	URL string `json:"url"`
+	// Git refspec, for example, refs/heads/my-branch
+	Branch string `json:"branch,omitempty"`
```
brandonjbjelland, line 12: Should we have a default of `main` instead of `omitempty`?

```suggestion
	// kubebuilder:default:=main
	Branch string `json:"branch"`
```

internal/controller/dataset_controller.go
```diff
@@ -190,6 +190,7 @@ func (r *DatasetReconciler) authNServiceAccount(sa *corev1.ServiceAccount) error
 }
 
 func (r *DatasetReconciler) buildJob(ctx context.Context, dataset *apiv1.Dataset) (*batchv1.Job, error) {
+	// TODO re-use the same buildJob for notebook, model and dataset controller
```
brandonjbjelland, line 193: yes! 👏

internal/controller/model_controller.go
```diff
@@ -249,7 +257,9 @@ func (r *ModelReconciler) modelImage(m *apiv1.Model) string {
 	}
 }
 
-func (r *ModelReconciler) buildJob(ctx context.Context, model *apiv1.Model, sourceModel *apiv1.Model) (*batchv1.Job, error) {
+// TODO move buildJob to separate file and let it take a Container spec
+// that way it can work with any resource
```
brandonjbjelland, line 261: 👏

brandonjbjelland:
I know this is a throwaway but I bet resolving the test failures would also make for good learnings! 🐉

api/v1/common.go
brandonjbjelland: Actually, I bet if you omit the default branch will be selected regardless of if it's main or master

samos123: That's correct. It will use the default branch configured on the remote so we shouldn't set it in substratus

Issue #83, opened 2023-07-11
Design and Implement a Blob Storage Interface (BSI) library and minimal implementation for GCS url signing


brandonjbjelland:
edit: heavily revised from a CRD to a more lightweight library and gRPC service

## Summary

As discussed this morning, we need a way to create signed URLs in GCS. This concrete task can be generalized considerably and it's likely we'll want to extend this library or service to support more storage operations (e.g. blob storage contents on a path, copy contents of a storage path to a different bucket in a different provider or region, etc.). I don't want to generalize too far beyond storage use cases here (a full Cloud Resource Manager seems overkill). This issue describes the new service implementing a Blob Storage Interface (BSI), its integration points with the core substratus Kubernetes controller.

## Design details

1. Blob Storage Interface (BSI) and GCS implementation:
The BSI will be a protobuf describing request and response payloads and the services it exposes. Our BSI itself is an artifact local to this repo but not something one would deploy - it's just an interface definition. For the first implementation, we'll be supporting Google Cloud Storage (GCS).

2. Kubernetes Controller integration:
Over the longer run, BSI implementations will be separate gRPC services installed optionally by the user potentially from dedicated repos. We'll enable pattern by creating the GCS BSI as a local library, and ideally communicating with it over gRPC.

3. Deployment Options:

We have several options for deploying the BSI within the cluster:

  1. As a separate Deployment with its own Service (available across the clster)
  2. As a sidecar container within the same Pod as the controller.
  3. As a library within the controller code itself.

I'll start as a library under its own directory as `internal/gcsBlobStorageService`. Open to discussion here. Moving toward independent deployments makes more sense in multi-tenant environments.

## Tasks

* [x] Design the BSI with GCS URL signing as the first supported operation. Make sure to design the BSI to be extensible for supporting additional cloud providers and operations in the future.
* [x] implement the gcsBlockStorageService. expose via a service
* [x] * [x] Decide on the deployment strategy for the BSI and implement it.
* [ ] integrate the controller
* [ ] Thoroughly test the core and edge cases.
* [ ] Document the design, usage, and any limitations of the system.

This new system will help us automate and manage our storage resources more securely going forward. Let's discuss this design and identify any potential issues or improvements but I'll likely start in before EOD to get a better handle on the 

As follow up, the kubectl plugin should be extended to be the first client here. Creating a separate issue for that work.

## Open questions

[...]

PR #86, opened 2023-07-12
adds a substratus cloud interface and GCS implementation

brandonjbjelland:
## What is this change?

Adds a gRPC service definition and a GCS server implementation as an internal package. The `main` function could serve the grpc endpoint (if its a distinct container/deployment/service, some change would be needed) but for now we could call `CreateSignedURL` from the controller directly using a `bsi.CreateSignedURLRequest`.

## Why make this change?

Clients (specifically `kubectl open notebook`) need to request a place to upload a tar.gz of the current workspace to trigger docker rebuilds of notebook containers. This service facilitates those actions via a gRPC service.

## Notes

Trying to keep the scope of change here small. I'm only creating the service/library and I'll have a separate PR for the controller integration.

## Links

https://github.com/substratusai/substratus/issues/83
[Slack thread](https://computeclubworkspace.slack.com/archives/C05CM6ZJKQX/p1689123418864049)


internal/bsi/bsi.proto
```diff
@@ -0,0 +1,21 @@
...
+package bsi.v1;
+option go_package = "github.com/substratusai/substratus/internal/bsi";
+
+service Controller {
+  rpc CreateSignedURL(CreateSignedURLRequest) returns (CreateSignedURLResponse) {}
+}
+
+message CreateSignedURLRequest {
+  string bucketName = 1;
+  string objectPath = 2;
```
brandonjbjelland, line 14: Originally I didn't have these but realized the controller is calling into the gcs library/service. We should expect the controller to have or build the details of `bucketName` and `objectPath`

brandonjbjelland: TODO in the controller integration PR - what bucket is most appropriate to use? Erring toward creating a dedicated bucket for notebook workspace uploads since this is neither a dataset or model.

internal/gcs/main.go
```diff
@@ -0,0 +1,76 @@
...
+	// Format the completion time to the proto spec
+	completionTime := timestamppb.New(time.Now())
+	if err != nil {
+		return nil, err
+	}
+
+	return &bsi.CreateSignedURLResponse{Url: url, CompletedAt: completionTime}, nil
+}
+
+func main() {
```
brandonjbjelland, line 56: I realize it's kind of nonsense to have a main function here creates a gRPC server since it can't actually be run directly (package is not `main`). Seems wrong to rename the package from `gcs` to `main` (while still living in the `internal/gcs` path mind you). A go-ism I don't know how to cope with just yet 💊

internal/gcs/main_test.go
brandonjbjelland, line 1: The effort of mocking and creating an interface for this code seems like a potential rabbit hole. If you want to see it, I'll make a pass at it in the morning.

brandonjbjelland: Cool - I'll delete it for the moment then

internal/gcs/main.go
```diff
@@ -0,0 +1,76 @@
...
+			// An error occurred that was NOT ErrObjectNotExist.
+			// This is an unexpected error and we should return it.
+			return nil, err
+		}
+		// the object doesn't exist and we want to continue and create the signed URL.
+	}
+
+	// Create a signed URL
+	url, err := storage.SignedURL(req.GetBucketName(), req.GetObjectPath(), &storage.SignedURLOptions{
+		Method:  "GET",
```
nstogner, line 40: Should this be POST so files can be uploaded?

internal/gcs/main.go
```diff
@@ -0,0 +1,76 @@
...
+		Expires: time.Now().Add(time.Duration(req.GetExpirationSeconds()) * time.Second),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// Format the completion time to the proto spec
+	completionTime := timestamppb.New(time.Now())
```
nstogner, line 48: Why include completion time as a response value?

internal/gcs/main.go
nstogner: Since you have already implemented a gRPC server, I would say lets just go forward with implementing this as a separate Deployment. I would move main into `$repo/cmd/gcpmanager/`. I would also move existing `$repo/cmd/main.go` into `$repo/cmd/controllermanager/main.go`.

internal/bsi/bsi.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
+
+import "google/protobuf/timestamp.proto";
+
+package bsi.v1;
```
nstogner, line 5: Considering that we will probably need to interact with other services like GAR, I wonder if it makes sense to split these proto pkgs out by the cloud service they are managing or just make a generic one like: `sci`: `Substratus Cloud Interface`. I wouldnt want to go all micro-servicey here... I think that its OK to keep this as `bsi` and just have a single server process implement multiple gRPC service interfaces... but have a slight preference for the combined (`sci`) gRPC service interface.

internal/gcs/main.go
nstogner: I just took a gander as the csi interface and they dont have this... I vote to remove it unless your have a compelling reason for keeping it.

internal/bsi/bsi.proto
nstogner: I think we either create an `uploads` bucket, or we create a bucket for each kind and use a subdirectory.

internal/gcs/main.go
samos123: I think the use case Brandon was thinking of is to know if the signed URL you get isn't expired already? I am not sure that's required though.

internal/bsi/bsi.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
```
nstogner, line 1: Is it possible to add the installation of the proto generation tool into the Makefile like the other tools (the local `bin/`)? If its not easy (i.e. if its not a go based tool), perhaps we could create a tooling Dockerfile.

nstogner: The invoking of the proto generation should probably be a part of `make generate`

internal/gcs/main.go
```diff
@@ -0,0 +1,76 @@
...
+	bsi "github.com/substratusai/substratus/internal/bsi"
+
+	storage "cloud.google.com/go/storage"
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+// server implements the bsi.ControllerServer interface.
+type server struct {
+	bsi.UnimplementedControllerServer
```
nstogner, line 20: This line should probably go away right?

internal/bsi/bsi.proto
brandonjbjelland: Ahh good call - both install and proto generation aren't the most straightforward. Adding that now.

internal/bsi/bsi.proto
brandonjbjelland: I started that path initially but it felt like the start of boiling the ocean to build an interface whose scope is all APIs of all providers. I think confining it to substratus cloud interface is a good way to keep the scope well understood and under control.

internal/gcs/main.go
brandonjbjelland: Good catch!

internal/gcs/main.go
brandonjbjelland: Can do. I'll do some shuffling...

internal/bsi/bsi.proto
brandonjbjelland: notebooks don't currently have a bucket. I think it makes a good case for creating one.

brandonjbjelland: renaming `objectPath` to `objectName` which looks more conventional

internal/gcs/main.go
brandonjbjelland: I don't think I have a great use case. If too short a window is requested by a client, they should have a reasonable idea of when the expiration is because it issued the request. Clients can still implement logic to handle retry under an assumed expiration in those case.

internal/gcs/main.go
brandonjbjelland: My understanding is this helps with compatibility. If you add a new method to the UnimplementedControllerServer interface, your existing server will still run without necessarily needing to add those methods. Comments in the generated code are terse but allude to this:

```
// ControllerServer is the server API for Controller service.
// All implementations must embed UnimplementedControllerServer
// for forward compatibility
```
and
```
// UnimplementedControllerServer must be embedded to have forward compatible implementations.
```

internal/gcpmanager/main.go
```diff
@@ -0,0 +1,51 @@
...
+		Scheme:  storage.SigningSchemeV2,
+		Method:  "POST",
+		Expires: time.Now().Add(time.Duration(req.GetExpirationSeconds()) * time.Second),
+	}
+
+	// Create a signed URL
+	// TODO(bjb): Need to test this on a live pod with WI with the SA having roles/iam.serviceAccountTokenCreator.
+	// I've only used this API with a static credential previously: https://cloud.google.com/storage/docs/access-control/signing-urls-with-helpers#upload-object
+	// a few techniques here: https://stackoverflow.com/questions/62439257/how-to-generate-signed-urls-for-google-cloud-storage-objects-in-gke-go
+	url, err := storage.SignedURL(bucketName, objectName, opts)
```
brandonjbjelland, line 45: Some testing to do here. I wonder if this will work. I really don't want to ship a static credential but the SO solution gives a reasonable alternative.

cmd/controllermanager/main.go
brandonjbjelland, line 1: I didn't find any references to the previous path so nothing else changed to make this move but I could have missed something.

internal/gcpmanager/main.go
brandonjbjelland: Indeed, this fails under a SA that has `iam.serviceAccountTokenCreator`:
```
failed to create signed URL: storage: missing required GoogleAccessID
```

Going to try the SO suggestion

cmd/gcpmanager/main.go
```diff
@@ -0,0 +1,58 @@
...
+		StorageClient: storageClient,
+	}
+
+	fmt.Println("calling CreateSignedURL with payload:")
+	resp, err := serv.CreateSignedURL(context.Background(), &payload)
+	if err != nil {
+		log.Fatalf("failed to create signed URL: %v", err)
+	}
+	fmt.Printf("signed URL: %v\n", resp.Url)
+}
```
brandonjbjelland, line 58: will remove before merge

internal/gcpmanager/main.go
```diff
@@ -0,0 +1,109 @@
...
+	// Create a signed URL
+	url, err := storage.SignedURL(bucketName, objectName, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	return &sci.CreateSignedURLResponse{Url: url}, nil
+}
+
+func getServiceAccountEmail() (string, error) {
```
brandonjbjelland, line 79: potentially a simpler path here: https://pkg.go.dev/golang.org/x/oauth2/google#FindDefaultCredentials

nstogner: Most google clients can be initialized like:

```go
		compute, err := computev1.NewService(context.Background())
		if err != nil {
```

which will find default credentials automatically

nstogner: Which should work out of the box with workload identity... also you should be able to make it work with some gclouding locally.

Makefile
```diff
@@ -180,3 +181,15 @@ $(EMBEDMD): $(LOCALBIN)
...
+
+.PHONY: protogen
+protogen: protoc ## Generate protobuf files.
+	cd internal/sci ; protoc --go_out=. --go_opt=paths=source_relative --go-grpc_out=. --go-grpc_opt=paths=source_relative sci.proto
+
+.PHONY: protoc
+protoc: $(PROTOC) ## download and install protoc.
+$(PROTOC): $(LOCALBIN)
+	test -s $(LOCALBIN)/protoc || \
+		brew install protobuf && \
```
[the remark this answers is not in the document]
brandonjbjelland: Yeah I like that idea.

Makefile
```diff
@@ -180,3 +202,41 @@ $(EMBEDMD): $(LOCALBIN)
...
+.PHONY: build_gcp_installer
+build_gcp_installer: ## Build the GCP installer.
+	docker build ./install -t substratus-installer
+
+.PHONY: gcp_installer_up
+gcp_installer_up: build_gcp_installer ## invoke the GCP installer to build all infra.
+	docker run -it \
+		-v ${HOME}/.kube:/root/.kube \
+		-e PROJECT=$(shell gcloud config get project) \
+		-e TOKEN=$(shell gcloud auth print-access-token) \
```
samos123, line 233: for dev I generally add this:
```
-e INSTALL_OPERATOR=no
```
that prevents the operator from being installed such that I can run make dev locally

internal/gcpmanager/main.go
brandonjbjelland: This is working now 🕺

Makefile
brandonjbjelland: Gotcha. I tidied this up so the long and repetitive docker command is captured in a variable and just the final bits are manipulated. `make gcp_installer_up_dev` is now your new friend!

config/gcpmanager/bootstrapper-job.yaml
brandonjbjelland, line 1: I haven't tested these yet 😬

install/terraform/gcp/buckets_iam.tf
```diff
@@ -1,11 +1,3 @@
-# Container Builder #
-
-resource "google_project_iam_member" "container_builder_gar_repo_admin" {
```
brandonjbjelland, line 3: replaced this with the more specific `artifact_registry_iam_membership`

install/terraform/gcp/buckets_iam.tf
```diff
@@ -50,3 +42,30 @@ resource "google_storage_bucket_iam_member" "data_loader_datasets_storage_admin"
   member = "serviceAccount:${google_service_account.data_loader.email}"
 }
 
+# GCP manager #
+
+resource "google_storage_bucket_iam_member" "gcp_manager_models_storage_admin" {
+  bucket = google_storage_bucket.models.name
+  role   = "roles/storage.admin"
```
brandonjbjelland, line 49: For the moment, `roles/storage.admin` is over-permissioning but we've already identified use cases where storage.admin will eventually be needed anyway (model or dataset deletion, signed urls for other reasons upload or download)

install/terraform/gcp/cluster.tf
```diff
@@ -93,7 +93,6 @@ resource "google_container_cluster" "main" {
       maximum       = 10480
     }
     # from https://cloud.google.com/compute/docs/gpus/#nvidia_gpus_for_compute_workloads
-    # TODO(bjb): ideally set these to align with regional quota on the project
```
brandonjbjelland, line 96: I think this is better handled elsewhere. Set these value to infinity

cmd/controllermanager/main.go
brandonjbjelland: Found lots of these later on when executing the test suite!

config/gcpmanager/bootstrapper-job.yaml
brandonjbjelland: now they work

internal/gcpmanager/main.go
brandonjbjelland: Maybe I haven't dug into this deeply enough but I wonder if the compute client actually gives what I'm after here... Is there any example you have of querying the metadata service but falling back on static creds?

I really just need the SA email address of the credentail being used. Maybe the simplest way to do that is to invoke different functions based on the presence of a static credential (SA key).

Makefile
```diff
@@ -180,3 +210,38 @@ $(EMBEDMD): $(LOCALBIN)
...
+### GCP installer targets
+
+RUN_SUBSTRATUS_INSTALLER := docker run -it \
+	-v ${HOME}/.kube:/root/.kube \
+	-e PROJECT=$(shell gcloud config get project) \
+	-e TOKEN=$(shell gcloud auth print-access-token) \
+	substratus-installer
+
+.PHONY: build_gcp_installer
+build_gcp_installer: ## Build the GCP installer.
```
nstogner, line 234: I think this should be `build_installer` ? PS: I would prefer to use `kebab-case` for Makefiles.

config/gcpmanager/gcp-manager.yaml
```diff
@@ -0,0 +1,56 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: gcp-manager
+  namespace: default
```
nstogner, line 5: The `gcp-manager` is a part of the Substratus control plane and should live in the `substratus` namespace.

docs/development.md
```diff
@@ -5,13 +5,7 @@
...
 
 ```sh
-docker build ./install -t substratus-installer && \
-
-docker run -it \
-  -v $HOME/.kube:/root/.kube \
-  -e PROJECT=$(gcloud config get project) \
-  -e TOKEN=$(gcloud auth print-access-token) \
-  substratus-installer gcp-up.sh
+make gcp_installer_up
```
nstogner, line 8: Perhaps `make install CLOUD=gcp DISABLE_CONTROLLER=true` and move the current `make install` to be `make install-crds`.

docs/development.md
```diff
@@ -64,10 +58,7 @@ kubectl apply -f examples/facebook-opt-125m/finetuned-model.yaml
 Cleanup.
 
 ```sh
-docker run -it \
-  -e PROJECT=$(gcloud config get project) \
-  -e TOKEN=$(gcloud auth print-access-token) \
-  substratus-installer gcp-down.sh
+make gcp_installer_down
```
nstogner, line 61: Perhaps `make uninstall CLOUD=gcp`

internal/gcpmanager/main.go
```diff
@@ -0,0 +1,108 @@
...
+	StorageClient *storage.Client
+}
+
+// CreateSignedURL generates a signed URL for a specified GCS bucket and object path.
+func (s *Server) CreateSignedURL(ctx context.Context, req *sci.CreateSignedURLRequest) (*sci.CreateSignedURLResponse, error) {
+	bucketName, objectName := req.GetBucketName(), req.GetObjectName()
+
+	bucket := s.StorageClient.Bucket(bucketName)
+	obj := bucket.Object(objectName)
+	if _, err := obj.Attrs(ctx); err != nil {
```
nstogner, line 29: `; err !=nil && err != storage.ErrObjectNotExist {`

internal/gcpmanager/main.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+		// the object doesn't exist and we want to continue and create the signed URL.
+	}
+
+	saEmail, err := getServiceAccountEmail()
+	if err != nil {
+		return nil, err
+	}
+
+	c, err := credentials.NewIamCredentialsClient(ctx)
```
nstogner, line 43: Typically, I like to follow the practice of instantiating all dependencies like clients in `main()` and storing them on the struct as a field `s.CredentialsClient`. This clearly shows the reader of `main()` what is being depended on. It also minimizes unnecessary calls as a part of each RPC invocation.

internal/gcpmanager/main.go
```diff
@@ -0,0 +1,108 @@
...
+	url := "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/email"
+
+	req, err := http.NewRequest(http.MethodGet, url, nil)
+	if err != nil {
+		fmt.Printf("Error creating request: %v\n", err)
+		return "", err
+	}
+	req.Header.Add("Metadata-Flavor", "Google")
+
+	client := &http.Client{}
```
nstogner, line 88: See comment about clients above.

internal/gcpmanager/main.go
```diff
@@ -0,0 +1,108 @@
...
+	url, err := storage.SignedURL(bucketName, objectName, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	return &sci.CreateSignedURLResponse{Url: url}, nil
+}
+
+func getServiceAccountEmail() (string, error) {
+	url := "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/email"
```
nstogner, line 79: Use pkg `cloud.google.com/go/compute/metadata`.

internal/gcpmanager/main.go
```diff
@@ -0,0 +1,108 @@
...
+	if _, err := obj.Attrs(ctx); err != nil {
+		if err != storage.ErrObjectNotExist {
+			// An error occurred that was NOT ErrObjectNotExist.
+			// This is an unexpected error and we should return it.
+			return nil, err
+		}
+		// the object doesn't exist and we want to continue and create the signed URL.
+	}
+
+	saEmail, err := getServiceAccountEmail()
```
nstogner, line 38: Currently this sends a call to the metadata server every time this RPC endpoint is invoked. This should be changed to do this once and store the service account email on the Server struct.

nstogner (changes requested):

Makefile
brandonjbjelland: Sgtm

config/gcpmanager/gcp-manager.yaml
brandonjbjelland: I hadn't considered that but spot on

internal/gcpmanager/main.go
brandonjbjelland: I like that. Can do

docs/development.md
brandonjbjelland: I'll take one of the two suggestions here and make `DISABLE_CONTROLLER` a flag. Let's leave the `CLOUD` argument until we need it and it does something meaningful 👍

docs/development.md
brandonjbjelland: Also pushing this one off until we have another platform we support. I'm all for it at that point!

internal/gcpmanager/main.go
brandonjbjelland: Ahh good call.

internal/gcpmanager/main.go
brandonjbjelland: Done

internal/gcpmanager/main.go
brandonjbjelland: FIgured this out with the metadata package. thanks for the pointer there 🙏

.github/workflows/docker-build-push.yaml
```diff
@@ -3,48 +3,62 @@ run-name: ${{ github.actor }} triggered Docker pipeline 🚀
 on:
   push:
     branches:
-    - main
+      - main
```
brandonjbjelland, line 6: excuse my formatter. As long as we don't have competing formatters, this is fine by me. If we do have competing formatters we can adopt a `.editorconfig`

nstogner (approved):
Looks great

PR #87, opened 2023-07-12
Falcon 7b instruct updates

samos123:
* Fixed Notebook reconciler to add DATA_PATH env variable when dataset is used
* Update falcon-7b-instruct example manifests


nstogner (approved):
Looks good. Just noticed a bug in the training code that will likely not allow the trainer to work: #88.

PR #90, opened 2023-07-13
fix #88 base models are not mounted

### LINKED ISSUE #88: Base Models are not mounted when training Models
Looks like `var baseModel` will always be nil... The culprit:

https://github.com/substratusai/substratus/blob/4bf967c6e98c57e41f4dff9f6b005e8e0e6e27eb/internal/controller/model_controller.go#L86

This line should be:

`baseModel = &apiv1.Model{}`

samos123:


brandonjbjelland (approved):
Easy stuff! LGTM

PR #94, opened 2023-07-14
Add cpu notebook

samos123:


brandonjbjelland (dismissed):
nice, simple, and cheap! 😛

examples/notebooks/cpu-notebook.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: substratus.ai/v1
+kind: Notebook
+metadata:
+  name: cpu-notebook
+spec:
+  suspend: false
+  image:
+    git:
+      url: https://github.com/substratusai/notebook-image
```
nstogner, line 9: Can we change thing to `notebook-minimal`? I am guessing we will publish a few variants.

nstogner (changes requested):

examples/notebooks/cpu-notebook.yaml
samos123: I was going to add a notebook-gpu-image afterwards that's based on the cuda base image.

Why don't I call it `notebook-minimal-cpu`?

samos123: done

nstogner (approved):

Issue #98, opened 2023-07-14
How to handle Model and Dataset restore from existing GCS contents?


brandonjbjelland:
## Summary

Our users may face a situation where their Kubernetes cluster is deleted, or the Custom Resource (CR) data gets lost due to unforeseen circumstances. While the underlying model and dataset data is preserved in storage, the users would not be able to point CRs at them to inform the controller of their existence. This issue is searching for a restore process, as users should not need need to rebuild datasets from scratch or retrain models when the underlying data to create those CRs is locally (in blob storage) available.

## Potential solutions

1. Extend our existing CRDs to allow importing directly from a storage path.
2. Write loader containers that import existing data and optionally diff against some source of truth (comparing hashes)

I lean toward prescribing solution 2 while keeping our ears open to feedback that solution 1 would be preferred. If we agree on that, this issue becomes an effort to create those data loader images.

## Links

[Slack thread](https://computeclubworkspace.slack.com/archives/C05CM6ZJKQX/p1689215981318749)

Issue #95, opened 2023-07-14
Consider including MD5 in the request to create a signed URL


brandonjbjelland:
In the near future, notebook clients will be able to request signedURLs for any reason, without any upper bound expiration duration, and with no restrictions on the upload contents. This is suboptimal and may give opportunities for a malicious user to man-in-the-middle the process.

One attribute in the [signing request options](https://pkg.go.dev/cloud.google.com/go/storage#SignedURLOptions) allows you to submit an MD5 of the contents. The godoc here speaks to the behavior for a GET method but I would think this also applies to the PUT method. The CRD would likely define this `spec.image.upload.md5`.

brandonjbjelland:
This is resolved

Issue #61, opened 2023-06-30
Publish base Substratus Dockerfiles


nstogner:
There could be a few base Substratus Dockerfiles that are published for Models and Datasets, with variations for different python versions (perhaps as tags).

These Dockerfiles might contain:

- cuda libraries
- standard scripts (i.e. `train.sh`, etc).
- Jupyter Labs pre-installed
- non-root directory permissions

nstogner:
Started working on this: https://github.com/substratusai/base-images

NOTE: This should also help with build and load times since base images layers will be more consistent.

brandonjbjelland:
This is done since we moved to the dedicated images repo 🥳

PR #100, opened 2023-07-14
Add proposed storage scheme & update model contract

nstogner:
Attempting to solve for:

- Simplified cluster migrations
- Consistent UX mounts across model training and notebook containers
- Better Jupyter UX: currently `/data` is outside of the file-explorer sidebar
- Reduce number of buckets needed for substratus (everything goes in one bucket)


docs/design.md
```diff
@@ -38,13 +38,112 @@ spec:
...
+All mount points will are made under a standardized `/ml` directory which should correspond to the `WORKDIR` of a Dockerfile. This works well for Jupyter notebooks which can be run with `/ml` set as the serving directory: all relevant mounts will be populated on the file explorer sidebar.
+
+The `logs/` directories below are used to store the notebook cell output, python logs, tensorboard stats, etc. These directories are mounted as read-only in Notebooks to explore the output of other background jobs.
+
+##### Dataset (importing)
+
+```
+/ml/params.json  # Populated from .spec.params (also available as env vars).
+
+/ml/data/        # Mounted RW: initially empty dir to write new files
```
nstogner, line 99: Should we use `dataset/` instead of `data/` to match `lower(.kind)`?

docs/design.md
```diff
@@ -38,13 +38,112 @@ spec:
...
+Storage scheme:
+
+```
+# Models
+gs://{project_id}-substratus-storage/namespaces/{namespace}/models/{name}/model
+gs://{project_id}-substratus-storage/namespaces/{namespace}/models/{name}/logs
+
+# Datasets
+gs://{project_id}-substratus-storage/namespaces/{namespace}/datasets/{name}/data
+gs://{project_id}-substratus-storage/namespaces/{namespace}/datasets/{name}/logs
```
brandonjbjelland, line 64: I knew S3 had issues when naming paths sequentially but wasn't sure on GCS. This [GCP blog post](https://cloud.google.com/blog/products/gcp/optimizing-your-cloud-storage-performance-google-cloud-performance-atlas) calls out upload performance impact to paths having the same or similar prefixes. Unsure if that applies to downloads.

We could apply a hashing scheme across a string of `"{project_id}-{namespace}-{name}"` as a `{computed_hash}` prefix and likely still be able to query using paths like the following to get all datasets in the bucket for a specific namespace:
```
gs://{project_id}-substratus-storage/**/namespaces/{namespace}/datasets/ *
```

Maybe this is premature optimization but at scale this def matters.

docs/design.md
```diff
@@ -38,13 +38,112 @@ spec:
...
+```
+
+NOTE: With this approach, a single bucket could be used to store all Models and Datasets for a given cluster.
```
brandonjbjelland, line 67: I'm trying to come up with a reason why the separation would be a good thing. Regulation for models has obviously not landed in a meaningful way anywhere but for customer data (and by extension, datasets), GDPR and friends are a consideration. Even within that regulatory scheme, we'd probably be fine mashing everything together.

I think I overall hedge toward keeping the bucket separation just because the uniform bucket access and IAM isolation makes permissioning specific SAs a little cleaner. Just my take!

docs/design.md
```diff
@@ -38,13 +38,112 @@ spec:
...
+}
+
+url := "{bucket}/namespaces/{namespace}/models/{name}"
+if sci.bucketObjectExists(url) {
+  # Use the bucket as the source of truth if .status.url did not exist.
+  updateStatus(url)
+  return
+}
+
+# ... Looks like the model should be imported/trained/etc.
```
brandonjbjelland, line 83: yeah I like this a lot better than my proposals 👏

docs/design.md
brandonjbjelland: I for one like that adjustment. Better consistency

docs/design.md
nstogner: I think we will need to go granular with the permissions over time… eek, so that’s why I think single might not matter

brandonjbjelland:
Resolves https://github.com/substratusai/substratus/issues/98

docs/design.md
```diff
@@ -38,13 +38,112 @@ spec:
...
+```
+
+##### Model (training)
+
+```
+/ml/params.json        # Populated from .spec.params (also available as env vars).
+
+/ml/data/              # Mounted RO: from .spec.trainingDataset
+
+/ml/saved-model/       # Mounted RO: from .spec.baseModel
```
samos123, line 119: shouldn't this be /ml/base-model

samos123: I see why it's /ml/saved-model because then it can be consistent with notebook

docs/design.md
nstogner: Updated PR to including hashing.

docs/design.md
```diff
@@ -38,13 +58,164 @@ spec:
...
 Optionally you can override the default command of a container by providing
 `command` in the Substratus resource:
 ```yaml
 spec:
   command: ["train.sh"]
 ```
 
+### User
+
+All containers should use the `ml` user for all training, importing, notebook, etc operations. The `/home/ml` directory should correspond to the `WORKDIR` of all Dockerfiles.
```
samos123, line 71: This would require updating all images and retesting, which I think is fine and rather do it earlier than later. Just making a note here to make sure this happens.

docs/design.md
```diff
@@ -38,13 +58,164 @@ spec:
...
+
+Non-root users should be enforced using the `runAsNonRoot: true` field within Pod security contexts.
```
samos123, line 73: I do think the usability of Substratus is impacted for folks that need to run custom images. For some folks writing a Dockerfile is hard enough when using root and having to use a specific user might make it too hard for them.

I am not entirely sold that the security benefits outweigh the UX downgrade here.

docs/design.md
```diff
@@ -38,13 +58,164 @@ spec:
...
+
+### Storage
+
+#### Buckets
+
+In Kubernetes the combination of the following fields make up a unique reference to a given object (i.e. database key): `split(.apiVersion, "/")[0] + .kind + .metadata.namespace + .metadata.name`. By storing related artifacts using a similar scheme, restore operations are made trivial: the administrator will not need to worry about backing up the `.status.url` field (as opposed to a scheme where storage location is `.metadata.uid` based - which would be unique for a given in-time instance of an object). When deleting and recreating clusters, if the storage bucket persisted, objects (i.e. Models, Datasets) can simply be re-applied into the new cluster (i.e. [velero](https://velero.io/)/similar is not needed). This plays nicely with GitOps.
```
samos123, line 79: nice! clever approach. Like it.

docs/design.md
```diff
@@ -38,13 +58,164 @@ spec:
...
+
+Pseudo-reconciler logic for a Model:
+
+```
+if .status.url != "" {
+  return
+}
+
+url := "{bucket}/{hash}"
+if sci.bucketObjectExists(url) {
```
samos123, line 149: would this check if the directory in bucket is present and has files? or just if directory is present?

docs/design.md
```diff
@@ -38,13 +58,164 @@ spec:
...
+}
+
+# ... Looks like the model should be imported/trained/etc.
+
+runJob()
+```
+
+#### Mount Points
+
+All mount points will are made under a standardized `/home/ml` directory which should correspond to the `WORKDIR` of a Dockerfile. This works well for Jupyter notebooks which can be run with `/home/ml` set as the serving directory: all relevant mounts will be populated on the file explorer sidebar.
```
samos123, line 162: what if my notebook requires root access?

docs/design.md
```diff
@@ -191,7 +363,8 @@ kind: Notebook
 name: notebook-training-experiment
 spec:
   image:
-    upload: {} # This is how the plugin signals it wants to upload a directory for building.
+    upload:
+      checksum: 11ddbaf3386aea1f2974eee984542152 # This is how the plugin signals it wants to upload a directory for building.
```
samos123, line 367: would this be the checksum of the tar file?

samos123:
Also requires updating the container contract doc

docs/design.md
brandonjbjelland: +1

docs/design.md
brandonjbjelland: yeah, related issue: https://github.com/substratusai/substratus/issues/95

docs/design.md
nstogner: A few things that would make this churn take less effort: 1. Using a base substratus image, 2. Using a monorepo for our models.

nstogner: Im starting work on 1: https://github.com/substratusai/substratus/issues/61

nstogner: Created an issue to track / discuss number 2: https://github.com/substratusai/substratus/issues/106

nstogner: Updated to remove root requirement.

docs/design.md
nstogner: Removed requirement for root.

docs/design.md
nstogner: Good question: also another consideration: the logic as written could result in an incomplete Model being loaded into the new cluster as a ready Model

docs/design.md
nstogner: You get it now, hurray!

nstogner:
Updated container contract

docs/design.md
```diff
@@ -38,13 +58,166 @@ spec:
...
+All mount points will are made under a standardized `/content` directory which should correspond to the `WORKDIR` of a Dockerfile. This works well for Jupyter notebooks which can be run with `/content` set as the serving directory: all relevant mounts will be populated on the file explorer sidebar.
+
+The `logs/` directories below are used to store the notebook cell output, python logs, tensorboard stats, etc. These directories are mounted as read-only in Notebooks to explore the output of other background jobs.
+
+##### Dataset (importing)
+
+```
+/content/params.json  # Populated from .spec.params (also available as env vars).
+
+/content/data/        # Mounted RW: initially empty dir to write new files
```
samos123, line 173: /content/dataset or /content/trainingDataset?

docs/container-contract.md
```diff
@@ -0,0 +1,48 @@
...
+  model/          # Location to store the resulting model from loading or training.
+  saved-model/    # Location where a previously saved model will be mounted.
+```
+
+### Parameters
+
+Substratus provides params as a file (`/content/params.json`) and as environment variables to containers.
+
+For example, the below Substratus Model spec will pass `abc: xyz`
+to the container using the environment variable `PARAM_ABC=xyz`:
```
samos123, line 39: what's the reason for not converting it to PARAMS? this confused me and didn't feel natural that the letter S gets removed.

docs/design.md
```diff
@@ -38,13 +58,166 @@ spec:
...
+Calculate a hash.
+
+Hashes are used to [prevent bucket hot-spots](https://cloud.google.com/blog/products/gcp/optimizing-your-cloud-storage-performance-google-cloud-performance-atlas).
+
+As opposed to a random id, a hash is used to provide a deterministic pattern storing and finding artifacts. By including a cluster name, Substratus can optionally store artifacts across multiple clusters in a single bucket if a compelling use-case arises in the future.
+
+```sh
+hash_input = "clusters/{cluster}/namespaces/{namespace}/{resource_plural}/{name}"
+
+hash = sha256(hash_input)
```
samos123, line 94: I thought you changed this to md5?

samos123 (dismissed):

docs/container-contract.md
nstogner: NOTE: Terraform allows for specifying `*.tfvars` files which then translate into `TF_VAR_{name}` environment variables, kind of like our use case: https://developer.hashicorp.com/terraform/cli/config/environment-variables#tf_var_name

docs/design.md
nstogner: Updated to include a `completed.json` file check.

samos123 (approved):

docs/container-contract.md
brandonjbjelland: I think `PARAM` is fine. My only q: why have `PARAM_` at all instead of `XYZ=abc`?

docs/design.md
```diff
@@ -38,13 +58,168 @@ spec:
...
+# Models
+gs://{bucket}/{hash}/model # Model artifacts (*.pt, etc)
+gs://{bucket}/{hash}/logs  # Logs and converted notebooks
+
+# Datasets
+gs://{bucket}/{hash}/data # Model artifacts (*.pt, etc)
+gs://{bucket}/{hash}/logs # Logs and converted notebooks
+
+# Notebooks
+gs://{bucket}/{hash}/build/{md5-checksum}.tar # Uploaded build context
```
brandonjbjelland, line 122: I spent a moment thinking about this last night and .zip would be more cross-platform friendly. So far I think we're supporting Windows still

docs/design.md
```diff
@@ -38,13 +58,168 @@ spec:
...
+NOTE: The `saved-model/` directory is the same as the container for the Model object when `.baseModel` is specified. This allows for easy development of Model training code.
+
+```
+/content/params.json        # Populated from .spec.params (also available as env vars).
+
+/content/data/              # Mounted RO: from .spec.dataset
+/content/data-logs/         # Mounted RO: from .spec.dataset
+
+/content/saved-model/       # Mounted RO: from .spec.model
+/content/saved-model-logs/  # Mounted RO: from .spec.model
```
brandonjbjelland, line 212: These are useful to see how the model was trained or how the dataset was originally built yes?

brandonjbjelland (approved):

docs/design.md
brandonjbjelland: oh right the libs would be doing the compression anyway. Yeah tar.gz is it! 👍

docs/container-contract.md
nstogner: B/c lowercase env vars are weird

docs/design.md
nstogner: Yep, notebooks are for exploration and iteration. This allows users to trace back the origin of the model which is useful IMO.

PR #102, opened 2023-07-15
use prebuilt image for falcon-7b-instruct

samos123:


samos123:
merging because Nick gave a thumbs up on slack

PR #103, opened 2023-07-15
Update falcon 7b instruct to use prebuilt images

samos123:


samos123:
merging because this aligns with prior agreement to use prebuilt images to accelerate adoption

PR #109, opened 2023-07-18
Refactor to match current design doc

### LINKED ISSUE #114: Support for params.json
As a part of the new container contract, `/content/params.json` should be mounted from `.spec.params`.

nstogner:
* Refactor cloud switch conditions into interfaces within the `cloud` pkg
* Update all bucket URLs to use hashes
* Add support for `params.json` (fixes #114)


internal/controller/container_reconciler.go
```diff
@@ -96,24 +108,26 @@ func (r *ContainerImageReconciler) ReconcileContainerImage(ctx context.Context,
...
 
 			if time.Now().After(expirationTime) {
-				log.Info("The signed URL has expired. Clearing .Status.ImageURL")
-				// TODO(bjb): why doesn't this work?
+				log.Info("The signed URL has expired. Clearing status.")
 				obj.SetStatusImage(ssv1.ImageStatus{
 					UploadURL:   "",
 					Md5Checksum: statusMd5,
 				})
+				if err := r.Client.Status().Update(ctx, obj); err != nil {
```
nstogner, line 122: Might solve number 2 in #111

.gitignore
```diff
@@ -71,5 +71,8 @@ dist/
 
 # envsubst rendered templates
 gcpmanager-skaffold.yaml
+
+/controllermanager
+/gcpmanager
```
brandonjbjelland, line 76: won't this ignore directories like `cmd/controllermanager` and `cmd/gcpmanager` and their `internal/` counterparts?

Makefile
```diff
@@ -109,13 +109,13 @@ skaffold-dev-gcpmanager: protoc skaffold protogen render-skaffold-manifests ## R
 build: manifests generate fmt vet ## Build manager binary.
 	go build -o bin/manager cmd/controllermanager/main.go
 
-.PHONY: dev
+.PHONY: releases
 dev: manifests kustomize install-crds
```
brandonjbjelland, line 113: I don't quite understand this - why have a PHONY (i.e., [a phony make target that doesn't correspond to a file that changes](https://stackoverflow.com/a/2145605)) that doesn't match the following line for the target action? I guess I've never seen this pattern where they don't match.

api/v1/notebook_types.go
```diff
@@ -62,7 +62,7 @@ type NotebookStatus struct {
 	// Conditions is the list of conditions that describe the current state of the Notebook.
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 
-	// Upload contains details the controller returns from a requested signed upload URL.
+	// Image contains the status of the image. Upload URL is reported here.
 	Image ImageStatus `json:"upload,omitempty"`
```
brandonjbjelland, line 66: Shouldn't this match the others?

```suggestion
	Image ImageStatus `json:"image,omitempty"`
```

api/v1/server_types.go
```diff
@@ -28,7 +28,7 @@ type ServerStatus struct {
 	// Conditions is the list of conditions that describe the current state of the Server.
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 
-	// Upload contains details the controller returns from a requested signed upload URL.
+	// Image contains the status of the image. Upload URL is reported here.
 	Image ImageStatus `json:"upload,omitempty"`
```
brandonjbjelland, line 32: likewise:

```suggestion
	Image ImageStatus `json:"image,omitempty"`
```

internal/cloud/cloud.go
```diff
@@ -0,0 +1,96 @@
...
+type BucketMount struct {
+	BucketSubdir  string
+	ContentSubdir string
+}
+
+type MountBucketConfig struct {
+	Container string        // Example: trainer, loader
+	Name      string        // Example: model, model-saved, data
+	Mounts    []BucketMount // Example: model, data, logs
+	ReadOnly  bool
```
brandonjbjelland, line 88: Does this type correspond to a single fs mount or all of them on the container? `Mounts` has me think it captures all of them but `ReadOnly` should probably only apply to a single mount point and not all right?

internal/cloud/common.go
```diff
@@ -0,0 +1,51 @@
...
+	"path/filepath"
+	"strings"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Common struct {
+	ClusterName       string     `env:"CLUSTER_NAME" validate:"required"`
+	ArtifactBucketURL *BucketURL `env:"ARTIFACT_BUCKET_URL,noinit" validate:"required"`
+	RegistryURL       string     `env:"REGISTRY_URL" validate:"required"`
```
brandonjbjelland, line 16: This gave an idea here: https://github.com/substratusai/substratus/issues/112

internal/controller/container_reconciler.go
brandonjbjelland: tyvm!

internal/controller/container_reconciler.go
```diff
@@ -26,8 +27,18 @@ import (
...
 type ContainerizedObject interface {
-	object
+	client.Object
+
+	GetConditions() *[]metav1.Condition
+
+	SetStatusReady(bool)
+
+	GetStatusImage() apiv1.ImageStatus
+	SetStatusImage(apiv1.ImageStatus)
```
brandonjbjelland, line 40: nit: Seeing that you defined `GetArtifactStatus` within types I would expect the getters and setters to be consistent

api/v1/model_types.go
```diff
@@ -44,6 +44,10 @@ func (m *Model) SetStatusReady(r bool) {
 	m.Status.Ready = r
 }
 
+func (m *Model) GetArtifactsStatus() ArtifactsStatus {
+	return m.Status.Artifacts
+}
```
brandonjbjelland, line 49: Is there a pref to use this function over attribute refs like: https://github.com/substratusai/substratus/pull/109/files#diff-c07787f303bc8908555dfa43fe6118a2582f65095931159c74133f1287d69732R67 ?

brandonjbjelland (changes requested):
A few questions worth addressing but overall this is huge and like 95% there

.gitignore
samos123: I don't think so because he added `/` in front of it. See relevant SO: https://stackoverflow.com/questions/24139478/when-to-use-leading-slash-in-gitignore

I didn't know about this either. The only reason I looked this up is because you asked in this PR :)

docs/design.md
```diff
@@ -112,12 +112,12 @@ The following scheme can be used for storing artifacts for Models, Datasets, and
 
 ```sh
 # Models
-gs://{bucket}/{hash}/model # Model artifacts (*.pt, etc)
-gs://{bucket}/{hash}/logs  # Logs and converted notebooks
+gs://{bucket}/{hash}/model     # Model artifacts (*.pt, etc)
```
samos123, line 115: maybe we could standardize on `gs://{bucket}/{hash}/data` for both models and datasets? Both datasets and models can store arbitrary files in arbitrary formats. In the end it's storing the Model's data and storing the Datasets'  data.

I don't have a preference, just bringing it up in case that makes stuff easier.

samos123: Feel free to resolve without responding if you think nah.

internal/cloud/cloud.go
samos123: nice catch. I think ReadOnly needs to move to BucketMount type. Otherwise how would you express this?

model(readonly=False), model-saved(readonly=True), data(readonly=True)

The code further down makes me believe this might be a bug indeed: https://github.com/substratusai/substratus/pull/109/files#diff-a1c985bf4c51fa451b98fc1431485adc3910b38bf897053cd4a881323083008fR66

Makefile
nstogner: I have no idea where this came from

internal/cloud/cloud.go
nstogner: GCSFuse allows for specifying readonly on an entire volume. The reason why I have multiple `.volumes` being appended is: I wanted to respect the bucket url reported by a given object instead of assuming that all objects in a given cluster are stored in the same bucket. One nice side effect of this: a Dataset object could eventually be imported into a GKE cluster that references a dataset from S3.

internal/cloud/common.go
nstogner: Agreed, as the code is written now, `REGISTRY_URL` should take precedence over the auto-configured GCR URL.

internal/controller/container_reconciler.go
nstogner: You mean `GetArtifactStatus` should be renamed to `GetStatusArtifacts`? I agree with that.

api/v1/model_types.go
nstogner: Hrm, this method may be cruft left over from an interface that was removed. I will revisit.

docs/design.md
nstogner: I went back-n-forth on a similar take: each object using `{hash}/artifacts`. I decided against that b/c we might want to store different types of "data" or "artifacts" and I thought it made sense to split those things by their types in directories.

internal/cloud/cloud.go
nstogner: Should probably rename to `type ObjectMount struct`?

internal/controller/container_reconciler.go
brandonjbjelland: Yeah you got my drift 🙏

.gitignore
brandonjbjelland: Ahh thanks for the doublecheck there

api/v1/model_types.go
nstogner: This method should be used when the underlying struct is not available (i.e. when it is being passed via an interface)

Makefile
nstogner: Changed

internal/cloud/cloud.go
brandonjbjelland: Thanks for the explanation - makes good sense now!

samos123 (dismissed):

brandonjbjelland (approved):

samos123:
merging and then tagging and releasing v0.5.0

PR #117, opened 2023-07-21
use prebuilt images in examples

samos123:


samos123:
merging after testing

Issue #84, opened 2023-07-11
Design and Implement a kubectl Plugin for the GCS Blob Storage Interface


brandonjbjelland:
## Summary
Let's extend the existing kubectl plugin to act as the first client of our GCS Blob Storage Interface (BSI) API. This kubectl plugin will help developers interact with the BSI API more conveniently, direct from the command line.

## Details

As proposed, [BSI](https://github.com/substratusai/substratus/issues/83) is an interface with exactly one implementation on GCS to begin. `GetSignedUrl` is the only service for now.

We could either extend the existing kubectl plugin or create a net-new one that handles BSI-related operations. This plugin will enable users to request signed URLs through a familiar UX.

## Tasks

* [x] Design the user interface for the kubectl plugin. What's the command's name? Flags? Defaults?
* [x] Implement the plugin.
* [x] Package and distribute the plugin as a standalone binary, similar to kubectl-open-notebook. It might make sense to collapse these into a single plugin. Let's discuss.
* [x] test the plugin with various scenarios and edge cases.
* [x] Document the installation process, usage, and limitations.

nstogner:
I would imagine that users would not be interfacing with these APIs, but rather they would actuate them through the existing CRDs... I.e. `kubectl open notebook` would interface with the Notebook API to request and upload URL be provided. It would then send a POST to the upload URL provided by the controller via the Notebook status.

brandonjbjelland:
Resolved by: https://github.com/substratusai/substratus/pull/147

PR #118, opened 2023-07-21
Feat: added a standalone upload resource plugin that works for all types

brandonjbjelland:
## What is this change?

Adds a dedicated kubectl plugin for upload capabilities across all resources that use `spec.image.upload.md5checksum`

## Why make this change?

We wanted to be able to exercise the signed url upload functionality through a kubectl plugin as our first client. Since the functionality exists across all our CRDs, I started (and finished) the path of making this flexible to all kinds. I've only tested against Notebooks so far and it's extremely fast with a dir that only contains a Dockerfile.

## Testing

When trying to run all the components in-cluster, things were broken and I didn't get the bottom of it. I opted to run both the GCP SCI and controller locally, it worked like a charm!

I was using the following Dockerfile to test (also included in the actual tests):

```
FROM ubuntu:18.04

RUN apt-get update -y && \
    apt-get install -y python3 && \
    mkdir api && echo foo> api/bar.txt

CMD [ "python3", "-mhttp.server", "--bind=0.0.0.0", "8888" ]
```

Executing like:
```
kubectl upload Notebook foobar -v -p ../minimal
```

Tar + upload takes a half second. The build job, once cached takes 7 seconds and a pod (if a node is ready), is live shortly thereafter. I stopped short of putting a watcher on the build job or the pod that should come up live if all goes correctly.

## Handling notebooks opening

I wonder if we'd be okay with one plugin calling another (I don't love creating a dependency between them) and just adding what's needed to finish out the task in `open notebook`. Maybe there's a more elegant way using the shared functions as a library. Open to thoughts here. I wanted to get this in front of you both before figuring out the specifics of handling notebooks.


brandonjbjelland:
resolves: https://github.com/substratusai/substratus/issues/84

kubectl/upload/main.go
```diff
@@ -0,0 +1,212 @@
...
+func run(cfg Config, client KubernetesClient) error {
+	if !fileExists(filepath.Join(cfg.Path, "Dockerfile")) {
+		return errors.New("a Dockerfile does not exist at the given path")
+	}
+
+	if cfg.Verbose {
+		fmt.Println("packaging the directory into a tarball")
+	}
+
+	tarPath := "/tmp/archive.tar.gz"
```
brandonjbjelland, line 107: Hardcoded this but open to suggestions.

kubectl/upload/main.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	if cfg.Verbose {
+		fmt.Printf("Upload is complete. Waiting for the build job to complete.\n")
+	}
+	// TODO(bjb): should we put a watcher also on name-kind-container-builder or does the utility stop here?
+	// spin.Start()
+	// spin.Suffix = " Waiting for the build job to complete..."
+	// spin.Stop()
+	// TODO(bjb): if it's a notebook, wait for it to be ready, then open it in a browser
```
brandonjbjelland, line 173: Likely changing the print contents and removing these lines. I'm thinking the remainder of the notebook specific functionality belongs in open-notebook.

kubectl/upload/utils_test.go
brandonjbjelland, line 1: These probably aren't worth keeping - the underlying libraries are well tested already.

kubectl/upload/main.go
nstogner: Tar to `bytes.Buffer` (i.e. in mem) instead of file. HTTP request body can be set from the buffer.

nstogner: `tarGz(path string, w io.Writer) error`

kubectl/upload/main.go
```diff
@@ -0,0 +1,212 @@
...
+	}}
+}
+
+func handleWatchEvent(event watch.Event, cfg Config, tarPath string) error {
+	updatedResource := event.Object.(*unstructured.Unstructured)
+
+	// Retrieve the value of .status.upload.uploadURL
+	status, ok, err := unstructured.NestedMap(updatedResource.Object, "status", "upload")
+	if err != nil || !ok {
+		return nil
```
nstogner, line 201: Should this be `return err`?

kubectl/upload/util.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+
+		// Skip the root directory
+		if path == src {
+			return nil
+		}
+
+		header, err := tar.FileInfoHeader(info, info.Name())
+		if err != nil {
+			return err
```
nstogner, line 62: Add context to errors as you return them with `return fmt.Errorf("tar header: %w", err)`

kubectl/upload/util.go
```diff
@@ -0,0 +1,133 @@
...
+		if err := tarWriter.WriteHeader(header); err != nil {
+			return err
+		}
+
+		if info.Mode().IsRegular() {
+			file, err := os.Open(path)
+			if err != nil {
+				return err
+			}
+			defer file.Close()
```
nstogner, line 88: I think if this fails the write to the file will fail and the error will tossed into the ether.

kubectl/upload/utils_test.go
nstogner: If you got some tests, no problem keeping them

kubectl/upload/util.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	defer tarFile.Close()
+
+	gzWriter := gzip.NewWriter(tarFile)
+	defer gzWriter.Close()
+
+	tarWriter := tar.NewWriter(gzWriter)
+	defer tarWriter.Close()
+
+	return filepath.Walk(src, func(path string, info os.FileInfo, err error) error {
```
nstogner, line 50: Add comment/issue to respect `.dockerignore` (can happen later).

kubectl/upload/main.go
```diff
@@ -0,0 +1,212 @@
...
+
+func main() {
+	var cfg Config
+
+	var cmd = &cobra.Command{
+		Use:   "upload [kind] [name]",
+		Short: "Upload a resource of a given Kind and Name to be built as a substratus container image",
+		Args:  cobra.ExactArgs(2),
+		Run: func(cmd *cobra.Command, args []string) {
+			cfg.Resource.Kind = args[0]
```
nstogner, line 65: `kubectl` convetions is to accept lowercase: `kubectl get pods` vs `kubectl get Pods`. As long as this works, I think we are good.

nstogner (dismissed):
Left some comments. Nothing show stopping, feel free to merge and iterate. PS: I dont think we should have cross-plugin exec'ing... We should structure our `kubectl/` directory to allow for shared libraries.

kubectl/upload/main.go
samos123: what if someone accidentally has a huge file in their directory? if we do decide to tar to memory then we should have a max tar size in mind that gets enforced. I think using a temporary path is fine.

I would prefer to use a MkdirTemp: https://pkg.go.dev/os#MkdirTemp and pass empty string for dir. 
> If dir is the empty string, MkdirTemp uses the default directory for temporary files, as returned by TempDir. Multiple programs or goroutines calling MkdirTemp simultaneously will not choose the same directory. It is the caller's responsibility to remove the directory when it is no longer needed.

samos123: Code you could use:
```
	dir, err := os.MkdirTemp("", "substratus")
	if err != nil {
		log.Fatal(err)
	}
	defer os.RemoveAll(dir) // clean up

	tarPath := filepath.Join(dir, "archive.tar.gz")
```

kubectl/upload/main.go
nstogner: Actually: `if err !=nil { return err`

... And: `if !ok { return errors.New("no upload status")`
?

nstogner:
`I stopped short of putting a watcher on the build job or the pod that should come up live if all goes correctly.` --> This should be keyed off of the `.status.ready` field on the substratus object instead.

.goreleaser.yaml
```diff
@@ -18,6 +18,18 @@ builds:
       - arm64
       - arm
       - ppc64le
+  - main: ./kubectl/upload/
```
nstogner, line 21: I would prefer `kubectl build-remote` ... `kubectl upload` is too broad IMO and might conflict with a future ability to upload a Dataset itself rather than the container for loading a Dataset... and longer term this plugin should only be registered for the following to avoid taking up too much kubectl plugin real estate (we should not assume Substratus is the only thing the end user is using k8s for).

```
kubectl build-remote model
kubectl build-remote server
kubectl build-remote notebook
kubectl build-remote dataset
```

kubectl/upload/main.go
brandonjbjelland: I initially started down the path of writing to memory but verifying things manually was more straightforward with a local file. I think I'll go back that direction now that I've seen it work with files on disk.

Memory vs disk - I think it'd probably be a mistake if the user wanted to upload and build a container image (specifically a loader, notebook, server) and it were so big that it didn't fit into memory (eg 15Gb+). Also, an upload that large has other places it might fail - for the moment I hardcoded the expiration time on the upload URL to 300 seconds. I'll verify what exactly this applies to this in follow up, but now that we automate the client-side but if that's the window for the entire upload, really large files might fail on slow connections anyway.

I would guess expiration duration actually applies to the initiation of the HTTP PUT and not the full duration of the upload but it's worth verifying. If it refers to initiation of the HTTP request, I bet we can safely shorten the window to 10 seconds.

samos123: I think it's fairly common in ML world to just use huge container images. That's how Substratus started out originally as well. So oI actually think using disk is quite important.

kubectl/upload/util.go
brandonjbjelland: Doy! Thank you

kubectl/upload/main.go
brandonjbjelland: Yeah, I was anticipating this watch will not have a `status.upload` immediately so I thought it was the right move to `return nil` but I'll spend some time here figuring out the correct behavior.

kubectl/upload/main.go
```diff
@@ -0,0 +1,212 @@
...
+			return errors.New("the custom resource was deleted")
+		default:
+			return errors.New("unhandled event type")
+		}
+	}
+
+	if cfg.Verbose {
+		fmt.Printf("Upload is complete. Waiting for the build job to complete.\n")
+	}
+	// TODO(bjb): should we put a watcher also on name-kind-container-builder or does the utility stop here?
```
brandonjbjelland, line 169: I'm going to follow through on this last bit because it applies to any resource that we're building. I think if a build fails, it's worth surfacing that to the client so it's not a complete mystery what happened on the in-cluster build.

kubectl/upload/main.go
brandonjbjelland: Ahh yeah it felt a bit strange using `Notebook` on the command line. I'll fix that up 👍

kubectl/upload/main.go
brandonjbjelland: Happy to follow a strong opinion here! 👍 tempdir incoming

kubectl/upload/main.go
nstogner: Ahh, ok that makes sense

kubectl/upload/main.go
nstogner: This info should be surfaced from the `.status.condtions` set by the container image reconciler

.goreleaser.yaml
brandonjbjelland: Yeah, I should have commented that I wasn't totally satisfied with `kubectl upload` - it's just too generic. The plugin started as `upload-kind` and became `upload-resource` prior to the current iteration.

Naming is hard but `build-remote` describes the high-level intent of the process better than any of the above. I'll go with that 👍 

I was actually okay punting on a less-than-ideal name here because medium-term I think we should get away from kubectl plugins and migrate toward a `substratus` or `ssai` CLI that rolls in these utilities. I'll capture that as an issue.

brandonjbjelland: Captured here: https://github.com/substratusai/substratus/issues/120

kubectl/upload/util.go
brandonjbjelland: I would think the build job itself would respect the `.dockerignore` but not compressing/uploading files in `.dockerignore` would be an improvement. 👍

kubectl/upload/util.go
brandonjbjelland: I'll add some output to make that more visibile.

brandonjbjelland:
Here's the output I'm getting from successful runs:

```
$ k delete notebook foobar ; sleep 2; cd kubectl/build-remote ; go build . ; cd - && kubectl/build-remote/build-remote notebook foobar -v -p ../minimal

notebook.substratus.ai "foobar" deleted
~/Development/git/github.com/substratusai/substratus
packaging the directory into a tarball
Created custom resource "foobar".
Upload URL is ready.
successfully uploaded tarball
Upload was successful. Waiting for the build job to complete.
Build job is complete and successful.
```

Occasionally (maybe 1/10) I get these while watching the status condition for build status. We're not handling them currently but we could if they persist:

```
2023/07/21 15:46:00 watch error occurred: an error on the server ("unable to decode an event from the watch stream: http2: response body closed") has prevented the request from succeeding
```

internal/controller/container_reconciler.go
```diff
@@ -281,6 +281,7 @@ func (r *ContainerImageReconciler) gitBuildJob(ctx context.Context, obj Containe
 
 	// Add an init container that will clone the Git repo and
 	// another that will append tini to the Dockerfile.
+	// TODO(any): we lost the tini init container here. On purpose?
```
brandonjbjelland, line 284: @nstogner JFYI

kubectl/build-remote/main.go
```diff
@@ -0,0 +1,301 @@
...
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+var (
+	ErrStatusNotFound  = errors.New("status not found")
+	ErrBuildIncomplete = errors.New("build not completed")
+	lastUploadURL      string
+)
```
brandonjbjelland, line 30: this doesn't feel conventional but I know a better option.

kubectl/build-remote/main.go
```diff
@@ -0,0 +1,301 @@
...
+		return fmt.Errorf("failed to watch resource %q: %w", customResource.GetName(), err)
+	}
+
+	for event := range watcher.ResultChan() {
+		switch event.Type {
+		case watch.Added, watch.Modified:
+			err = handleWatchEvent(event, cfg, "build")
+			if errors.Is(err, ErrBuildIncomplete) {
+				// When build is complete, stop the watcher
+				time.Sleep(500 * time.Millisecond)
```
brandonjbjelland, line 209: These sleeps seem to limit hitting this error class

PR #139, opened 2023-07-24
add custom prompt model trainer example

samos123:


examples/falcon-7b-instruct/finetuned-model-custom-prompt.yaml
```diff
@@ -0,0 +1,27 @@
...
+  trainingDataset:
+    name: k8s-instructions
+  params:
+    epochs: 1
+    prompt_template: |
+      You're a helpful assistant that helps with generating K8s YAML manifests. Your responses should only be YAML or multi YAML files
+      ## Instruction
+      {prompt}
+      ## Response:
+      {completion}
```
nstogner, line 23: I think we should leave off the `{completion}` bit. That way we can re-use this template for Server types.

nstogner: It can be implicit that completion follows the end of the prompt template.

nstogner (changes requested):

examples/falcon-7b-instruct/finetuned-model-custom-prompt.yaml
samos123: I don't think that's needed. In the server it would simply not fill in the completion and that would be totally fine. I rather keep it as-is instead of having implicit things which might confuse end-users. It's not a given that completion should always be at the end

samos123: discussed with Nick we can revisit this later once we work on server, but current plan for server would be to strip anything after {completion}

nstogner (approved):

brandonjbjelland (approved):
Lgtm

PR #146, opened 2023-07-26
add llama2-7b and llama2-70b examples

samos123:


examples/llama2-7b/base-model.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: substratus.ai/v1
+kind: Model
+metadata:
+  name: llama-2-7b
+spec:
+  image: substratusai/model-loader-huggingface
+  params:
+    name: meta-llama/Llama-2-7b-hf
+    hugging_face_hub_token: ${HUGGINGFACE_TOKEN}
```
nstogner, line 9: Its the user supposed to replace this before applying? If so, we should have a comment here.

samos123: yes I was using this:
```
export HUGGINGFACE_TOKEN=mahtoken
cat examples/llama2-70b/base-model.yaml | envsubst | kubectl apply -f -
```

I will make a comment in the file itself and write these steps in user doc

samos123: I added a comment, please let me know if that's good enough.

examples/llama2-7b/base-model.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: substratus.ai/v1
```
brandonjbjelland, line 1: Adding a comment to the top of the file to drive the point home
```suggestion
# ensure this manifest is rendered via envsubst and $HUGGINGFACE_TOKEN  is set prior to apply

apiVersion: substratus.ai/v1
```

examples/llama2-7b/base-model.yaml
brandonjbjelland: suggestino added above: https://github.com/substratusai/substratus/pull/146/files#r1285113274

examples/llama2-7b/base-model.yaml
brandonjbjelland: Or feel free to use the same format/position as the comment in the 70b manifest: https://github.com/substratusai/substratus/pull/146/files#diff-c15d5042193b36d46847eba0cac2a61c593da60e386243bc5ae2aaf60a9c3acaR9-R12

brandonjbjelland (approved):
🍍

examples/llama2-7b/base-model.yaml
samos123: I rather keep it next to where HUGGINGFACE_TOKEN is used. Or you mean to have 2 comments? 1 at the top and 1 next to where the variable is used?

samos123: The reason being is some people might just want to copy paste directly in the file instead of relying on `envsubst`. I just gave that as an example on how you could replace it.

PR #149, opened 2023-07-28
add fine tuned example with new params

samos123:


brandonjbjelland (approved):
Yep makes good sense to align with the HF API

PR #150, opened 2023-07-30
set save_steps to 5 by default

samos123:
The HuggingFace transformer.TrainingArguments uses a default value of 500, however that doesn't make sense for our k8s dataset which only uses 71 steps. So we should set a smaller save_steps value in our examples.

This also helps with users understanding how to override specific parameters.

### DIFF
--- a/examples/falcon-7b-instruct/finetuned-model-custom-prompt.yaml
+++ b/examples/falcon-7b-instruct/finetuned-model-custom-prompt.yaml
@@ -11,7 +11,7 @@ spec:
     name: k8s-instructions
   params:
     num_train_epochs: 1
-    # Save to checkpoint every 5 steps
+    # Save to checkpoint every 5 steps for a dataset with ~70 steps total.
     save_steps: 5
     prompt_template: |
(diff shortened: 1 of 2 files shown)


examples/falcon-7b-instruct/finetuned-model-custom-prompt.yaml
```diff
@@ -11,6 +11,8 @@ spec:
     name: k8s-instructions
   params:
     num_train_epochs: 1
+    # Save to checkpoint every 5 steps
+    save_steps: 5
```
brandonjbjelland, line 15: Adding a little to the comment to help contextualize why you chose 5 for the given dataset:

```suggestion
    # Save to checkpoint every 5 steps for a dataset with ~70 steps total.
    save_steps: 5
```

brandonjbjelland (approved):
One suggestion but LGTM otherwise

samos123:
added your suggestion and will merge if CI passes

brandonjbjelland (approved):

PR #154, opened 2023-08-01
add gitattibutes file

d3287t328:
Just some basic file definitions and enabling of github features.


samos123:
it seems all tooling we use already understands the correct file format based on the extension. What's the use case for adding this .gitattributes file? Trying to understand what the benefit of it is.

d3287t328:
> it seems all tooling we use already understands the correct file format based on the extension. What's the use case for adding this .gitattributes file? Trying to understand what the benefit of it is.

it is a code quality enhancement for Github that will help for things like Github Actions if anyone uses it downstream.

d3287t328:
Closed by request.

PR #161, opened 2023-08-02
Puts docker-login action behind a conditional - allows outside contributors to pass CI

brandonjbjelland:
## What is this change?

Adds conditionals to docker-login, allowing outside contributors to get successful builds on their PRs.

## Why make this change?

As we get outside contributions, it'd be great to see our entire build pipeline succeed on their PRs. This change allows us to skip the action that requires access to secrets.

## Related

An example: https://github.com/docker/metadata-action/tree/v4/
https://github.com/substratusai/substratus/pull/154


.github/workflows/docker-build-push.yaml
```diff
@@ -7,9 +7,25 @@ on:
     tags:
       - "v*.*.*"
   pull_request:
+  workflow_dispatch:
+  issue_comment:
+    types: [created]
 jobs:
   docker-pipeline:
     runs-on: ubuntu-latest
+    if: |
```
nstogner, line 16: Can you add a comment as to why the conditionals are all happening here instead of using the filters in the `on:` block at the top? Perhaps a link to the inspiration for this? How does this approach compare to other well-known OSS projects?

.github/workflows/docker-build-push.yaml
```diff
@@ -7,9 +7,25 @@ on:
...
+        (github.event_name == 'push' && (
+          github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v')
+          )
+        ) ||
+        github.event_name == 'pull_request' ||
+        github.event_name == 'workflow_dispatch' ||
+        (github.event_name == 'issue_comment' &&
+        github.event.issue.pull_request &&
+        contains(github.event.comment.body, '/test') &&
+        (github.event.comment.author_association == 'MEMBER' ||
```
nstogner, line 26: Are we attempting to implement AuthZ logic in here? Is this tackled already within GitHub?

nstogner: PS: Does this solve our needs? https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

nstogner (changes requested):

.github/workflows/docker-build-push.yaml
brandonjbjelland: Sure.

Spelling it out here quick (since I prefer the soln below if it works) - In short, the `on:` block is necessary but not sufficient to allow these alternate trigger methods. At best the comment-based-trigger is not useful to an outside contributor because like the PR trigger, credentials won't be passed properly. At worst, it's a security risk to allow anyone but contributors and maintainers to trigger the flow - I think secrets become available to the workflow under this context and they can be echoed.

This pattern seems reasonably well-established. Here's the [github doc](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#accessing-secrets) outlining the risk in tokens leaking

Conditional pattern examples:
https://stackoverflow.com/questions/75079250/how-to-trigger-new-job-by-comment-in-pull-request-for-github-actions
https://github.com/orgs/community/discussions/25389

.github/workflows/docker-build-push.yaml
brandonjbjelland: Let me give this workflow run a try. I'd prefer that if it works since the 'managed' workflow would actually show up in the PR UI (and failing would block) unlike how it works on a comment based trigger

brandonjbjelland:
It looks like this is the only known path forward: https://iterative.ai/blog/testing-external-contributions-using-github-actions-secrets

But also, I ultimately think we don't need this - just PRs don't need to push to dockerhub at all, so they shoulnd't hit the login job. The conditional should live at a different place and block the job from proceeding beyond building the container image. We don't need container images to publish as a PR is submitted right? If we do, we can gate it behind a similar contributor/maintainer check like the above.

brandonjbjelland:
@nstogner - I revised this to be simpler and smaller. All we needed was to put the action that uses secrets behind a conditional and they won't be accessed during PRs (saves us a step too). The docker build action will still run and not push as expected. Reference: https://github.com/docker/metadata-action/tree/v4/

samos123 (approved):

PR #162, opened 2023-08-02
Add live-syncing from notebook to local machine

### LINKED ISSUE #140: Feature Request: Notebook support constant file sync
Projects like ksync can be considered to implement constant filesysnc instead of current mechanism where sync only happens during initial creation and once you hit Ctrl + C. Ksync itself probably can't be used as-is but the same mechanism should work.

Issues with current approach:
* If the pod gets deleted for some reason it will cause you to lose all your modifications
* Unable to view the changes in real-time in your local notebook viewer (e.g. vscode)

nstogner:
Fixes #140 

TODO: Arch description and diagrams.


containertools/cmd/nbwatch/main.go
```diff
@@ -0,0 +1,78 @@
...
+func run() error {
+	w, err := fsnotify.NewWatcher()
+	if err != nil {
+		return err
+	}
+	defer w.Close()
+
+	w.Add("/content/src")
+
+	go watchLoop(w)
```
samos123, line 29: why does `watchLoop` need to be started in a goroutine? I see it has an endless for loop in it anyway and then you could get rid of the block forever line?

samos123 (dismissed):

containertools/cmd/nbwatch/main.go
```diff
@@ -0,0 +1,78 @@
...
+		// Read from Errors.
+		case err, ok := <-w.Errors:
+			if !ok { // Channel was closed (i.e. Watcher.Close() was called).
+				return
+			}
+			klog.Error(err)
+		// Read from Events.
+		case e, ok := <-w.Events:
+			if !ok { // Channel was closed (i.e. Watcher.Close() was called).
+				return
```
samos123, line 48: should we log the event or will `e` be null?

samos123:
This scenario might cause a user to lose their local changes:
1. In your local directory make some changes  to the src/my-book.ipynb
2. Create a notebook that uses a prebuilt image
3. Use kubectl notebook to sync local directory

Assumption of what would happen, the notebook sync would sync the src/my-book.ipynb from the prebuilt image to my local directory. That would overwrite my local changes.

Workaround: Ensure only sync local directory can be used when building the image from local directory?

nstogner:
@samos123 Agreed, that scenario could be an issue. A tricky one to figure out a solution for, but a good one to keep noodling on.

containertools/cmd/nbwatch/main.go
nstogner: Thats a good catch, I ripped this from the fsnotify example. I think in our case, we could avoid the goroutine.

containertools/cmd/nbwatch/main.go
nstogner: In this if block I think `e` is its empty value.

samos123:
I don't think that scenario should block this PR. I can file a separate issue for tackling that scenario.

samos123 (approved):

brandonjbjelland (approved):
🚀

PR #166, opened 2023-08-03
remove downloading SA key

samos123:
Many GCP orgs have strict policy that prevents downloading SA key. I thought about using impersonation but that also doesn't seem very convenient and would change the developers config when using something like this:
```
gcloud config set auth/impersonate_service_account SERVICE_ACCT_EMAIL
```

This changes back to original way of working where the active default credentials of gcloud are used.

I validated this works for me when using Project Owner as the currently authenticated user in gcloud

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,9 +126,6 @@ dev-up:
 	  -e TF_VAR_attach_gpu_nodepools=false \
 	  -e INSTALL_OPERATOR=false \
 	  substratus-installer gcp-up.sh
-	mkdir -p secrets
-	gcloud iam service-accounts keys create --iam-account=substratus-gcp-manager@$(shell gcloud config get project).iam.gserviceaccount.com ./secrets/gcp-manager-key.json
-
 .PHONY: dev-down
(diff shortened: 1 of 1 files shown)


nstogner:
`make dev-run` works and serves both the SCI and the controller is working?

nstogner:
I will test this on my machine as well and hit approve if it works

nstogner:
The gcp-manager fails silently on my machine when you remove the env var / key:

```sh
go run ./cmd/gcpmanager & \
        go run ./cmd/controllermanager/main.go \
                --sci-address=localhost:10080 \
                --config-dump-path=/tmp/substratus-config.yaml
Your active configuration is: [my-project-here]
2023/08/03 14:56:02 failed to get the SA email: GOOGLE_APPLICATION_CREDENTIALS environment variable not set
exit status 1
```

samos123:
It did work for me. So let me check my default env variables

samos123:
This indeed breaks gcpmanager, so closing for now

PR #169, opened 2023-08-05
add artifact URL as soon as reconciling starts

### LINKED ISSUE #144: Provide GCS URL once model is created instead of once ready
Right now the GCS url is only provided once the model is ready, however it would be nice to be able to pull down the logs as the model is being loaded/trained. So hoping that if we provide the URL I can pull down logs directly while it's loading.

samos123:
Fixes #144

### DIFF
--- a/api/v1/conditions.go
+++ b/api/v1/conditions.go
@@ -1,12 +1,11 @@
 package v1
 
 const (
-	ConditionReconciling = "Reconciling"
-	ConditionUploaded    = "Uploaded"
-	ConditionBuilt       = "Built"
-	ConditionLoaded      = "Loaded"
-	ConditionModelled    = "Modelled"
-	ConditionDeployed    = "Deployed"
+	ConditionUploaded = "Uploaded"
+	ConditionBuilt    = "Built"
+	ConditionLoaded   = "Loaded"
+	ConditionModelled = "Modelled"
+	ConditionDeployed = "Deployed"
 )
 
 const (
--- a/api/v1/conditions.go
+++ b/api/v1/conditions.go
@@ -1,11 +1,12 @@
 package v1
 
 const (
-	ConditionUploaded = "Uploaded"
-	ConditionBuilt    = "Built"
-	ConditionLoaded   = "Loaded"
-	ConditionModelled = "Modelled"
-	ConditionDeployed = "Deployed"
+	ConditionReconciling = "Reconciling"
+	ConditionUploaded    = "Uploaded"
+	ConditionBuilt       = "Built"
+	ConditionLoaded      = "Loaded"
+	ConditionModelled    = "Modelled"
+	ConditionDeployed    = "Deployed"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


internal/controller/model_controller.go
```diff
@@ -68,10 +68,23 @@ func (r *ModelReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 func (r *ModelReconciler) reconcileModel(ctx context.Context, model *apiv1.Model) (result, error) {
 	log := log.FromContext(ctx)
 
-	if model.Status.Artifacts.URL != "" {
+	if model.Status.Ready {
```
nstogner, line 71: Keep this change to the conditional

internal/controller/model_controller.go
```diff
@@ -68,10 +68,23 @@ func (r *ModelReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
...
 		return result{success: true}, nil
 	}
 
+	if model.Status.Artifacts.URL == "" {
+		model.Status.Artifacts.URL = r.Cloud.ObjectArtifactURL(model).String()
```
nstogner, line 76: Always set this URL.

internal/controller/model_controller.go
```diff
@@ -68,10 +68,23 @@ func (r *ModelReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
...
+		meta.SetStatusCondition(model.GetConditions(), metav1.Condition{
+			Type:               apiv1.ConditionReconciling,
+			Status:             metav1.ConditionTrue,
+			Reason:             apiv1.ReasonJobComplete,
+			ObservedGeneration: model.Generation,
+		})
+		if err := r.Status().Update(ctx, model); err != nil {
```
nstogner, line 83: Remove this status update... This will be too fleeting, not worth the API request.

nstogner: This update request will be sent very soon after this and the URL will be set then.

api/v1/conditions.go
```diff
@@ -1,11 +1,12 @@
 package v1
 
 const (
-	ConditionUploaded = "Uploaded"
-	ConditionBuilt    = "Built"
-	ConditionLoaded   = "Loaded"
-	ConditionModelled = "Modelled"
-	ConditionDeployed = "Deployed"
+	ConditionReconciling = "Reconciling"
```
nstogner, line 4: Conditions here should be end-states not transition states. IMO, I would remove this and go with the suggested changes below which make this unnecessary.

nstogner (changes requested):
Agree we should set URL earlier, but I don't think we want to include a new call to update status. I would just set the field and allow the next update status call to do what you are intending here.

internal/controller/model_controller.go
samos123: you mean change model.Status.Ready == true?

samos123: I think what I have currently is preferred over having
```
if model.Status.Ready == true {
```

Source: https://stackoverflow.com/questions/3685002/check-if-boolean-is-true

samos123:
I verified the new approach of not using an additional update works perfectly as well. Thanks for the great suggestion

internal/controller/model_controller.go
nstogner: Was just commented on checking ready instead of the url

nstogner (approved):
Looks good!

brandonjbjelland (approved):

internal/controller/model_controller.go
samos123: not sure why I misunderstood your comment. I read it as change to the conditional format or smth like that.

Issue #12, opened 2023-06-21
Support for AWS


nstogner:
To support AWS we would need to:

- Add `infra/terraform/aws`
- Add `infra/commands/{aws-commands}`
- Add `AWSCloudContext` to `CloudContext`
- Conditionally add AWS-specific settings (for example: IRSA annotations for k8s ServiceAccounts)
- Find an equivalent to GCSFuse for S3

brandonjbjelland:
I wanted to get a start on understanding GCSFuse equivalents. It turns out AWS makes it difficult or charges for the privilege of using S3 as a file system! S3 notably doesn't make an appearance on the [k8s-csi supported drivers list](https://kubernetes-csi.github.io/docs/drivers.html). Here's what I came away with:

1. [s3fs-fuse](https://github.com/s3fs-fuse/s3fs-fuse) seems to have a good amount of mindshare but doesn't support IRSA. This [linked project](https://github.com/efrecon/docker-s3fs-client) scratches at how one might use it in EKS but the production-readiness is def not there.  
2. yandex-cloud produces a [geesefs-based](https://github.com/yandex-cloud/geesefs) s3 filesystem with a [csi driver](https://github.com/yandex-cloud/k8s-csi-s3) having close to POSIX compliance. Some [support here](https://stackoverflow.com/questions/73841839/kubernetes-csi-s3-driver-with-readwritemany-in-2022) on SO. The geesefs readme (if it's to be believed) gives a nice [perf](https://github.com/yandex-cloud/geesefs) and [POSIX compatibility](https://github.com/yandex-cloud/geesefs#posix-compatibility-matrix) table comparing to other projects here. Of those, I don't actually think rclone fits the bill but potentially.
3. [goofys](https://github.com/kahing/goofys) looks compelling from a performance and cross-platform standpoint (a [HN post](https://news.ycombinator.com/item?id=35789225) also has a fan) but looks to have shortcomings compared to the two above. I wonder if the first non-POSIX behavior outlined is a dealbreaker or not:

* only sequential writes supported

Checkpointing models, logging nb outputs, saving datasets - I believe these are all constitute sequential write operations but it's hard to know for sure. This blog post outlines how to use it within an EKS context: https://dev.to/otomato_io/mount-s3-objects-to-kubernetes-pods-12f5

Users of s3fs-fuse seem to have in part migrated toward goofys due to lack of support.

[...]

PR #170, opened 2023-08-06
adds AWS infra for instantiating and destroying all baseline substratus components

brandonjbjelland:
## What is this change?

This change adds a script that we (and users) can use to create a complete substratus environment on a new AWS account.

## Why make this change?

https://github.com/substratusai/substratus/issues/12

## Caveats, questions, TBD

~Q: Are the local/ephemeral SSDs important for any sort of workload we support? Trying to understand if that's critical here too.~ - size is important but local SSDs probably not. Leaving this out for now.

Unknown: I added karpenter but haven't throughly tested how well it works at auto-provisioning. GPU backed noe

~Caveat: The features we get enabled on a GKE cluster through simple flags are **incredibly** fussy on EKS and I don't have them working. I'll dig further (maybe I've missed some events) but I think we might need to manage these on our own. e.g. I've never seen these fail on GKE:~

~Equivalent features baked into EKS configuration fail consistently. I've seen timeouts across `coredns`, `vpc-cni`, `ebs-csi` regardless of the order of deployment or how I do it. So far I've tried:
1. using the EKS module flags
2. breaking out into `eks_add_on` resources ([docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_addon)) and
3. using [a purpose-built module](https://registry.terraform.io/modules/aws-ia/eks-blueprints-addons/aws/1.5.0?utm_content=documentLink&utm_medium=Visual+Studio+Code&utm_source=terraform-ls).
 
They all fail. We need to attach some of the IRSA roles created in the code here to those resources however we instantiate them so if that's out of scope of terraform, some additional outputs will be necessary (i.e., arns of the roles).~

## With time, I'll add to this...

1. an equivalent to gcp-up.sh and gcp-down.sh ✅ 
2. an aws-up Dockerfile ✅ 
3. make targets ✅ 
4. docker hub push actions ✅  - not needed. rolled into same container image


BOsterbuhr:
Hey @brandonjbjelland and team, I had a quick thought I wanted to pass along so I don't forget about it.

> Need to explore further if this configuration will actually work or if it's node auto-provisioning is smart enough in this case. I suspect EKS is not going to make good instance type/accelerator choices on our behalf.

[Karpenter](https://karpenter.sh) may be a good option to get the intelligent auto-provisioning you are looking for. [Karpenter docs on provisioners](https://karpenter.sh/docs/concepts/provisioners/).
The [install](https://aws-ia.github.io/terraform-aws-eks-blueprints/v4.18.0/add-ons/karpenter/) _should_ be easy if you use the [blueprints addons module](https://aws-ia.github.io/terraform-aws-eks-blueprints-addons/main/) but as you mentioned the addons aren't always so straightforward. You could always do a helm install of Karpenter if the addon path doesn't work.

I'll double-check what I ended up doing to get `coredns`, `vpc-cni`, and `ebs-csi` working consistently in my terraform.

I am looking forward to testing Substratus out on Monday.

samos123:
>Are the local/ephemeral SSDs important for any sort of workload we support? Trying to understand if that's critical here too.

You got a point there, I don't see a critical need yet, so  we can skip them for now.

samos123:
Spinning up EKS clusters with Terraform is such a pain, maybe by design? I wasn't expecting it to be so complex. The eksctl tool does seem to make it easier:
https://eksctl.io/usage/gpu-support/
https://eksctl.io/usage/eksctl-karpenter/

Not sure if it will fit all our needs or takes away too much flexibility. I personally prefer terraform but seeing the struggle and complexity of EKS, it might be fine to consider something like `eksctl` for development. Note I have never tried `eksctl` myself so it could be even worse.

In the end we expect most users to already have a K8s cluster when they want to use Substratus and those end-users will choose their own tooling of choice to manage EKS + nodegroups. So the main purpose of the bundled installer/EKS cluster creator is mostly development and PoC phase to get rolling quickly with minimal issues.

brandonjbjelland:
@BOsterbuhr thanks for the pointer! Karpenter looks viable here and could very well be a common autoscaler we could use across other providers.

Sidebar: if you're hoping to use substratus on AWS, we're just getting started on adding support. Running on GCP is the paved path we have today.

BOsterbuhr:
In my experience if you are just trying to get dev/PoC support for AWS quickly then `eksctl` is good choice.

As for Karpenter, yeah unfortunately AWS doesn't seem to be in a rush to support other cloud providers.

Sidebar: I'll test on GCP first to get a better understanding of everything and will watch for AWS support.

brandonjbjelland:
> In my experience if you are just trying to get dev/PoC support for AWS quickly then `eksctl` is good choice.

We discussed internally today and arrived at this same consensus. In short, delivering a cluster is not our value add, where we should spend time, or put in a lot of code (which inevitably will rot, have feature requests itself, etc.). We just need the simplest possible way to get a minimum cluster up in each supported provider for someone starting at zero - that may or may not be through terraform. Here it seems `eksctl` is the right-sized approach. 👍 

> As for Karpenter, yeah unfortunately AWS doesn't seem to be in a rush to support other cloud providers.

Though we're being very careful with the dependencies we take on, I don't know that it changes our calculation here - karpenter def presents itself as the simplest way to auto-scale EKS on hydrogenous on hardware with minimal overhead. We very much feel incentivized to avoid having a long list of node groups for the different flavors of GPU-supported instances if that's our alternative.

> Sidebar: I'll test on GCP first to get a better understanding of everything and will watch for AWS support.

Thank you! 🙏  Any feedback is highly appreciated, @BOsterbuhr ! ❤️

install/kubernetes/karpenter-provisioner.yaml.tpl
```diff
@@ -0,0 +1,79 @@
...
+      value: "true"
+      effect: NoSchedule
+  requirements:
+    - key: karpenter.sh/capacity-type
+      operator: In
+      values: ["spot"]
+    - key: node.kubernetes.io/instance-type
+      operator: In
+      values:
+        # aws ec2 describe-instance-types --region us-west-2 --query "InstanceTypes[?GpuInfo!=null].InstanceType" --output json | jq -r '.[]' | sort | grep -v dl1 | grep -v inf | grep -v p5 | grep -v trn1 | awk '{print "\""$1"\","}'
```
BOsterbuhr, line 37: If you want you can use this instead of having to list them all out
```
- key: karpenter.k8s.aws/instance-category
          operator: In
          values: ["g", "p"]
- key: karpenter.k8s.aws/instance-family
          operator: NotIn
          values: ["p5"]
```

install/Dockerfile
```diff
@@ -1,36 +1,67 @@
...
   apt-get install -y \
   gnupg \
   software-properties-common \
   unzip \
   wget \
   curl \
-  git
+  git \
+  tzdata \
+  gettext-base \
```
brandonjbjelland, line 35: required because I'm using `envsubst` in the aws-up and aws-down

install/kubernetes/karpenter-provisioner.yaml.tpl
brandonjbjelland: This is a good simplification and one that we'd need to extend a little further to exclude graviton architecture hosts (our images aren't produced or tested against arm yet). I'll cook that up. Thanks!

install/Dockerfile
```diff
@@ -1,36 +1,67 @@
...
+  keyboard-configuration
```
samos123, line 36: what is the `keyboard-configuration` package used for? I  don't care if we include it but just sparkled my curiosity.

install/kubernetes/eks-cluster.yaml.tpl
```diff
@@ -0,0 +1,96 @@
...
+        # https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-presigned-url.html
+        Version: "2012-10-17"
+        Statement:
+          - Effect: Allow
+            Action:
+              - "s3:PutObject"
+              - "s3:GetObject"
+            Resource:
+              - "arn:aws:s3:::${ARTIFACTS_BUCKET_NAME}/*"
+              - "arn:aws:s3:::${ARTIFACTS_BUCKET_NAME}"
```
samos123, line 96: Let's tackle this later in a follow up discussion.

I think the K8s SA and identity mapping should not go into the cluster creation, that should stay a separate step. The main reason being the purpose of cluster creation should be to mimick what an end-user cluster might likely look like.

So the K8s SA + IRSA mapping for controller itself should be a separate install step using pure AWS commands.

install/kubernetes/eks-cluster.yaml.tpl
```diff
@@ -0,0 +1,96 @@
...
+          - Effect: Allow
+            Action:
+              - "s3:*"
+              - "s3-object-lambda:*"
+            Resource:
+              - "arn:aws:s3:::${ARTIFACTS_BUCKET_NAME}/*"
+              - "arn:aws:s3:::${ARTIFACTS_BUCKET_NAME}"
+    - metadata:
+        name: aws-manager
+        namespace: ${CLUSTER_NAME}
```
samos123, line 85: namespace should be hardcoded to be `substratus` for now

samos123 (changes requested):
The only change I would like to see before merging is hardcoding namespace to be substratus instead of using `$CLUSTER_NAME` because `substratus` is what K8s YAML manifest hardcodes as well.

This was much more pleasant to review than the terraform code to create an EKS cluster

install/kubernetes/eks-cluster.yaml.tpl
brandonjbjelland: This is for aws-manager, similar to gcp-manager. I think we had @nstogner on the fence whether or not it's a good idea to have the core controller `*-manager` wihtin the cluster and this is related but falls on the cloud provider API side of that install process. I thought we more or less decided cloud provider API actions mostly fell in the domain of an outside provisioner like tf or eksctl (for now anyway).

I think the concern with allowing the controller to create a SA, its IRSA trust policy binding, and assign it S3 permissions is how permissive we have to allow the controller to be (project-level) as opposed to it's cloud API scope to only be working against service account permissions via modifying trust policy bindings (and the GCP equivalent, workload identity bindings).

Perhaps we split the which entity does what role: the controller handles the ISRA binding but the outside provisioner does everything else (creates the role and the initial policy). Unsure.

install/Dockerfile
brandonjbjelland: I actually don't think it's needed - I found an aws cli install guide that listed it but I dont actually think it's necessary. Getting rid of it and tzdata.

install/kubernetes/eks-cluster.yaml.tpl
brandonjbjelland: The first IRSA listed here is definitely not what we decided on earlier in the week and I'll be changing it:

```yaml
    - metadata:
        name: ${CLUSTER_NAME} # formerly would have been hardcoded as container-builder
        namespace: ${CLUSTER_NAME} # should be hardcoded as substratus
      attachPolicy:
        Version: "2012-10-17"
        Statement:
          - Effect: Allow
            Action:
              - "ecr:*"
            Resource:
              - "arn:aws:ecr:::${ARTIFACTS_REPO_NAME}"
```

As I look at this more, I think the mega-provider-identity™️  we should be rolling up into **is** the `aws-manager` role or `gcp-manager` SA.

brandonjbjelland:
As a note here - I still haven't seen karpenter successfully launch a GPU-attached node for me though I see it's trying. Here's where it currently stands when I try to launch a pod specifying an accelerator:

```
# 2023-08-09T08:37:36.198Z    ERROR    controller    Reconciler error    {"commit": "34d50bf-dirty", "controller": "machine.lifecycle", "controllerGroup": "karpenter.sh", "controllerKind": "Machine", "Machine": {"name":"nvidia-gpu-rdlfj"}, "namespace": "", "name": "nvidia-gpu-rdlfj", "reconcileID": "9bd22025-a33d-4cd9-bba9-14957cf6967a", "error": "creating machine, creating instance, with fleet error(s), InvalidParameterValue: Value (eksctl-KarpenterNodeInstanceProfile-substratus) for parameter iamInstanceProfile.name is invalid. Invalid IAM Instance Profile name"}
```

For:

```
apiVersion: v1
kind: Pod
metadata:
  name: ubuntu-gpu
  namespace: default
  labels:
    app: ubuntu
spec:
  containers:
    - name: ubuntu
      image: ubuntu:latest
      command: ["/bin/sleep", "3650d"]
      imagePullPolicy: IfNotPresent
      resources:
        limits:
          nvidia.com/gpu: 1 # Requesting 1 GPU
  restartPolicy: Always
```

Makefile
```diff
@@ -120,40 +120,60 @@ skaffold-dev-gcpmanager: protoc skaffold protogen render-skaffold-manifests ## R
...
+		substratus-installer aws-down.sh
+
+.PHONY: gcp-dev-run
 # Controller manager configuration #
-dev-run: export CLOUD=gcp
-dev-run: export GPU_TYPE=nvidia-l4
-dev-run: export PROJECT_ID=$(shell gcloud config get project)
-dev-run: export CLUSTER_NAME=substratus
-dev-run: export CLUSTER_LOCATION=us-central1
+gcp-dev-run: export CLOUD=gcp
```
nstogner, line 168: Preference for `dev-*` pattern (perhaps `dev-run-gcp`). I also included this on the website repo.

install/Dockerfile
```diff
@@ -1,36 +1,65 @@
 FROM ubuntu:23.04
 
 WORKDIR /workspace
+# Determine platform and architecture
+RUN ARCH=$(uname -m) && \
```
nstogner, line 5: Slight preference for a `get-tools.sh` script for all tool-fetching bits and a call to `RUN get-tools.sh` in Dockerfile. The get-tools.sh script looks like it would be usable outside of the container, and it would keep the Dockerfile a little more organized.

install/scripts/aws-up.sh
```diff
@@ -0,0 +1,93 @@
...
+aws iam create-service-linked-role \
+  --aws-service-name spot.amazonaws.com || true
+
+aws eks update-kubeconfig \
+  --region ${REGION} \
+  --name ${CLUSTER_NAME}
+
+# Logout of helm registry to perform an unauthenticated pull against the public ECR
+helm registry logout public.ecr.aws || true
+helm upgrade \
```
nstogner, line 53: Full Karpenter install looks to be supported using eksctl from my reading. Any reason to install using helm?

`It will create all the necessary prerequisites outlined in Karpenter's [Getting Started](https://karpenter.sh/docs/getting-started/) section including installing Karpenter itself using Helm`

https://eksctl.io/usage/eksctl-karpenter/#karpenter-support

install/scripts/aws-up.sh
```diff
@@ -0,0 +1,93 @@
...
+  --set controller.resources.limits.memory=1Gi \
+  --wait
+
+envsubst <${KUBERENTES_DIR}/aws/karpenter-provisioner.yaml.tpl >${KUBERENTES_DIR}/aws/karpenter-provisioner.yaml
+kubectl apply -f ${KUBERENTES_DIR}/aws/karpenter-provisioner.yaml
+
+# node-termination-handler: https://artifacthub.io/packages/helm/aws/aws-node-termination-handler
+helm repo add eks https://aws.github.io/eks-charts
+helm upgrade \
+  --install aws-node-termination-handler \
```
nstogner, line 74: `We recommend against using Node Termination Handler alongside Karpenter due to conflicts that could occur from the two components handling the same events.`

https://karpenter.sh/docs/faq/#should-i-use-karpenter-interruption-handling-alongside-node-termination-handler

`If you're using [EKS managed node groups](https://docs.aws.amazon.com/eks/latest/userguide/managed-node-groups.html), you don't need the aws-node-termination-handler.`

https://github.com/aws/aws-node-termination-handler#project-summary

install/scripts/aws-up.sh
```diff
@@ -0,0 +1,93 @@
...
+# Required env variables:
+: "$AWS_ACCOUNT_ID $AWS_ACCESS_KEY_ID $AWS_SECRET_ACCESS_KEY"
+
+INSTALL_OPERATOR="${INSTALL_OPERATOR:-yes}"
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+KUBERENTES_DIR=${SCRIPT_DIR}/../kubernetes
+
+EKSCTL_ENABLE_CREDENTIAL_CACHE=1
+export CLUSTER_NAME=substratus
```
nstogner, line 15: Preference for following bash conventions of using lowercase variables within scripts:

```bash
cluster_name=substratus
```

https://stackoverflow.com/questions/673055/correct-bash-and-shell-script-variable-capitalization

Exception would be the variables that you are using `envsubst` with.

install/scripts/aws-up.sh
```diff
@@ -0,0 +1,93 @@
...
+export REGION=us-west-2
+export ARTIFACTS_REPO_NAME=${CLUSTER_NAME}
+export ARTIFACTS_BUCKET_NAME=${AWS_ACCOUNT_ID}-${CLUSTER_NAME}-artifacts
+export KARPENTER_VERSION=v0.29.2
+export AWS_PARTITION="aws"
```
nstogner, line 20: If including partition, lets add a link to what it is: https://docs.aws.amazon.com/whitepapers/latest/aws-fault-isolation-boundaries/partitions.html

I am not sure we need this for our purposes.

install/kubernetes/aws/eks-cluster.yaml.tpl
```diff
@@ -0,0 +1,102 @@
...
+metadata:
+  name: ${CLUSTER_NAME}
+  region: ${REGION}
+  version: "1.27"
+  tags:
+    createdBy: eksctl
+    environment: dev
+    karpenter.sh/discovery: ${CLUSTER_NAME}
+
+karpenter:
```
nstogner, line 12: Prefer to set minimal number of options:

```yaml
karpenter:
  version: 'v0.29.0'
```

nstogner (changes requested):

Makefile
brandonjbjelland: No disagreement from me here. Done.

install/Dockerfile
brandonjbjelland: Yeah can do

install/scripts/aws-up.sh
brandonjbjelland: I didn't read the whole way through the FAQ. Good catch. removing.

install/scripts/aws-up.sh
brandonjbjelland: I'm comfortable hardcoding it and not documenting

From: https://karpenter.sh/v0.29/getting-started/getting-started-with-karpenter/#2-set-environment-variables

install/kubernetes/aws/eks-cluster.yaml.tpl
brandonjbjelland: I don't have confidence karpenter would work without some of these.

1. `createServiceAccount` - without this I don't think we have a karpenter role. We'd need to create our own by some other process and inject it to karpenter configuration. I'll take this line over that effort.
2. `withSpotInterruptionQueue` - we could probably get rid of this but why? From the schema doc:
> if true, adds all required policies and rules for supporting Spot Interruption Queue on Karpenter deployments
Isn't that something we and customers would likely want? This is effectively the replacement component for the node-termination-handler [docs](https://aws.amazon.com/blogs/containers/using-amazon-ec2-spot-instances-with-karpenter/).
3. `defaultInstanceProfile` - has to be the same across both the provisioner and this manifest. I think we need it or we're guessing or relying on convention of those tools.
4. `version` - :shrug: we might be able to get away without? I think version guarantees are a good thing to ensure our users get something that is known to work.

install/scripts/aws-up.sh
brandonjbjelland: done.

install/scripts/aws-up.sh
brandonjbjelland: I see what you meant, we don't need the helm install if `eksctl` is doing that same thing for us 👍 Removing those bits.

install/kubernetes/eks-cluster.yaml.tpl
nstogner: `I think the concern with allowing the controller to create the mega-SA, its IRSA trust policy binding, and assign it S3/ECR permissions is how permissive we have to allow the controller to be (project-level)` --> I think Sam was actually saying the Role should be created in the installation step, but instead of with eksctl, using the aws cli.

install/scripts/aws-up.sh
```diff
@@ -0,0 +1,55 @@
...
+  --region ${REGION} >/dev/null || true
+
+envsubst <${kubernetes_dir}/aws/eks-cluster.yaml.tpl >${kubernetes_dir}/aws/eks-cluster.yaml
+eksctl create cluster -f ${kubernetes_dir}/aws/eks-cluster.yaml ||
+  eksctl upgrade cluster -f ${kubernetes_dir}/aws/eks-cluster.yaml
+
+aws iam create-service-linked-role \
+  --aws-service-name spot.amazonaws.com || true
+
+aws eks update-kubeconfig \
```
nstogner, line 34: I think you get this for free from `eksctl`?

install/scripts/aws-down.sh
```diff
@@ -0,0 +1,33 @@
...
+EKSCTL_ENABLE_CREDENTIAL_CACHE=1
+export CLUSTER_NAME=substratus
+export REGION=us-west-2
+export ARTIFACTS_REPO_NAME=${CLUSTER_NAME}
+export ARTIFACTS_BUCKET_NAME=${AWS_ACCOUNT_ID}-${CLUSTER_NAME}-artifacts
+
+(aws eks update-kubeconfig \
+  --region ${REGION} \
+  --name ${CLUSTER_NAME} &&
+  kubectl delete deployments --namespace=karpenter --all &&
```
nstogner, line 21: Any reason to delete these namespaces as opposed to blasting the cluster?

install/scripts/aws-down.sh
```diff
@@ -0,0 +1,33 @@
...
+  kubectl delete deployments --namespace=kube-system --all) ||
+  true
+
+envsubst <${kubernetes_dir}/aws/eks-cluster.yaml.tpl >${kubernetes_dir}/aws/eks-cluster.yaml
```
nstogner, line 25: I think you can blast the cluster away just by name without the config?

nstogner (approved):
Good stuff! Added some comments, but none of them are big deals

install/scripts/aws-down.sh
brandonjbjelland: If only! the cloudformation stacks would still be left over and EKS doesn't let you blow away a cluster without deleting node groups first.

install/scripts/aws-down.sh
brandonjbjelland: Weirdly, the cloudformation stack will find itself stuck and unable to actually delete the node groups due to unevictable pods in these namespaces. See: https://github.com/substratusai/substratus/pull/170#discussion_r1289325977 for why not just delete the cluster.

I suppose one approach might be deleting the namespaces. That would probably be a suitable alternative, esp if other resource types are blocking (daemonsets)

install/scripts/aws-up.sh
brandonjbjelland: Yes but I wonder if it becomes your context. Docs say:

> After the cluster has been created, the appropriate kubernetes configuration will be added to your kubeconfig file.

install/scripts/aws-down.sh
nstogner: Thanks for the context. Can you add a comment to describe the need so I don’t forget and try to “simplify” later?

samos123 (dismissed):

samos123:
Karpenter seems to have Karpenter specific node labels you have to use in your pod spec. This might require some more design discussion and require making use_karpenter a flag or expose it on Substratus resources somehow?

install/kubernetes/eks-cluster.yaml.tpl
samos123: Spot on Nick. That was exactly what I was trying to say. So someone that already has an existing cluster can easily follow along as well.

BOsterbuhr:
> Karpenter seems to have Karpenter specific node labels you have to use in your pod spec.

Can you explain what you mean? Your pods shouldn't have to even know Karpenter exists.

samos123:
Let's say you want to run a pod on A100 GPU then how would you ensure the pod gets scheduled on a node that has A100 GPU on Karpenter vs non Karpenter? You might have nodes with T4, V100 and A100.

Note I might be totally wrong since I haven't used Karpenter myself. I was reading this: https://karpenter.sh/preview/concepts/scheduling/

That doc made me believe in order for Karpenter to create a node that has A100 I would have to set nodeSelector in the pod to karpenter.k8s.aws/instance-gpu-name = a100 OR use affinity rules. 

I got a GCP background so this is my first time seriously looking into Karpenter.

BOsterbuhr:
> Let's say you want to run a pod on A100 GPU then how would you ensure the pod gets scheduled on a node that has A100 GPU on Karpenter vs non Karpenter? You might have nodes with T4, V100 and A100 in the same cluster.

Oh ok, that makes sense, so instead of making the end user understand that you are using Karpenter you could just create a provisioner with a [User-Defined Label](https://karpenter.sh/docs/concepts/scheduling/#user-defined-labels) as a requirement and then have your end-user use that label as a node selector.
Another requirement you would have then in that provisioner would be karpenter.k8s.aws/instance-gpu-name = a100
I am far from an expert in Karpenter but I believe you ideally want to keep most things at the provisioner level.

samos123:
The issue is that there doesn't seem to be a node label that exposes the GPU type on AWS unless you use Karpenter, however at the same time we also don't want to depend on Karpenter and ensure Substratus works well without Karpenter. A key principle of Substratus is to minimize dependencies so it's easier to get Substratus to run in any EKS cluster.

Actually, I might be wrong all together and should just get a GPU node on EKS to verify myself. Seems there is in fact a label that would have the info we're looking for: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go#L38C28-L38C31

BOsterbuhr:
For example you could do
```
apiVersion: karpenter.sh/v1alpha5
    kind: Provisioner
    metadata:
      name: a100
    spec:
      requirements:
        - key: karpenter.k8s.aws/instance-gpu-name
          operator: In
          values: ["a100"]
        - key: gpu-type/A100
          operator: Exists
        - key: "karpenter.sh/capacity-type"
          operator: In
          values: ["spot"]
      taints:
      - key: "nvidia.com/gpu"
        effect: NoSchedule
```
or
```
apiVersion: karpenter.sh/v1alpha5
    kind: Provisioner
    metadata:
      name: v100
    spec:
      requirements:
        - key: karpenter.k8s.aws/instance-gpu-name
          operator: In
          values: ["v100"]
        - key: gpu-type/V100
          operator: Exists
        - key: "karpenter.sh/capacity-type"
          operator: In
          values: ["spot"]
      taints:
      - key: "nvidia.com/gpu"
        effect: NoSchedule
```
then when creating the pods you would add either
```
nodeSelector:
    gpu-type/A100: "true"
```
or
```
nodeSelector:
    gpu-type/V100: "true"
```

BOsterbuhr:
> Actually, I might be wrong all together and should just get a GPU node on EKS to verify myself. Seems there is in fact a label that would have the info we're looking for

yeah you're right it looks like that is where the google label you mentioned is coming from as well https://github.com/kubernetes/autoscaler/blob/fc5870f8eaf850dd1e18a5884a7491168dc5d8a0/cluster-autoscaler/cloudprovider/gce/gce_cloud_provider.go#L37

The only issue I could see is one I think you all brought up previously; when using the Kubernetes autoscaler you have to manage a separate node group of each different instance type. But that may be worth it if you don't want any dependencies.

samos123 (approved):

install/kubernetes/eks-cluster.yaml.tpl
brandonjbjelland: I get it now! Issue created: https://github.com/substratusai/substratus/issues/178

PR #171, opened 2023-08-06
add proposal for Substratus managed infra

samos123:


docs/proposals/operator-managed-infra.md
```diff
@@ -41,7 +41,7 @@ environments without having to learn IaC tooling.
   gets called when a new Service Account is provisioned in a new namespace:
   ```
   gcloud iam service-accounts add-iam-policy-binding substratus@my-project.iam.gserviceaccount.com \
-   --role roles/iam.workloadIdentityUser \
+   --role roles/iam. workloadIdentityUser \
```
brandonjbjelland:
```suggestion
--role roles/iam.workloadIdentityUser \
```

docs/proposals/operator-managed-infra.md
```diff
@@ -78,7 +84,7 @@ Open Questions:
     --role "roles/storage.admin" --role "roles/artifactregistry.repoAdmin"
    ```
 
-   The role roles/iam.serviceAccountAdmin is needed to be able to use workload identity across multiple
+   The role `iam.serviceAccountAdmin`` is needed to be able to use workload identity across multiple
```
brandonjbjelland, line 87:
```suggestion
The role `iam.serviceAccountAdmin` is needed to be able to use workload identity across multiple
```

brandonjbjelland (approved):

docs/proposals/operator-managed-infra.md
```diff
@@ -0,0 +1,116 @@
...
+
+### Implementation details
+* Object storage: Create bucket if the configured bucket does not exist
+* Image Registry: Create registry if the configured registry does not exist
+* New sub reconciler for namespace: Create K8s Service Accounts automatically in each namespace where Substratus is used and
+  annotate the K8s Service Account with the correct GCP Service Account. Ensure that following
+  gets called when a new Service Account is provisioned in a new namespace:
+  ```
+  gcloud iam service-accounts add-iam-policy-binding substratus@my-project.iam.gserviceaccount.com \
+   --role roles/iam. workloadIdentityUser \
```
brandonjbjelland, line 44:
```suggestion
   --role roles/iam.workloadIdentityUser \
```

docs/proposals/operator-managed-infra.md
```diff
@@ -0,0 +1,116 @@
...
+* Image Registry: Create registry if the configured registry does not exist
+* New sub reconciler for namespace: Create K8s Service Accounts automatically in each namespace where Substratus is used and
+  annotate the K8s Service Account with the correct GCP Service Account. Ensure that following
+  gets called when a new Service Account is provisioned in a new namespace:
+  ```
+  gcloud iam service-accounts add-iam-policy-binding substratus@my-project.iam.gserviceaccount.com \
+   --role roles/iam.workloadIdentityUser \
+   --member "serviceAccount:myproject.svc.id.goog[new-namespace/substratus]"
+  ```
+  Make sure this only happens whenever a new Substratus gets created in a new namespace.
```
brandonjbjelland, line 47:
```suggestion
  Make sure this only happens whenever a new Substratus resource is created in a new namespace.
```

docs/proposals/operator-managed-infra.md
```diff
@@ -0,0 +1,116 @@
...
+  The controller should be smart enough to not have to create KSAs in namespaces that
+  do not have any substratus resource.
```
brandonjbjelland, line 49:
```suggestion
  do not have any substratus resources.
```

docs/proposals/operator-managed-infra.md
```diff
@@ -0,0 +1,116 @@
...
+
+How should Substratus controller handle infra management?
+* Terraform
+* Native K8s controller that uses Golang SDK for AWS/GCP/..
```
brandonjbjelland, line 53: strong preference for just using the go sdk within the controller.

docs/proposals/operator-managed-infra.md
```diff
@@ -0,0 +1,116 @@
...
+The Native K8s controller would better fit within the existing Substratus code base.
+It would also allow us to verify if a resource is already there and if not just create
+it. Terraform would still be used for creation of GKE/EKS cluster to provide a reference
+example for easy install.
+
+### Open Questions
+
+1. Where should the code to create Image Registry and Bucket live? Do we need a new
+   resource or can we reconcile based purely on configuration?
+2. Should Substratus install and manage Karpenter on AWS?
```
brandonjbjelland, line 64: I believe we landed on no for this and similar cluster-level DaemonSets (e.g., nvidia drivers) though cluster configurations like this would be good to verify through the controller. If incompatible configurations are found, the controller should likely emit warning level logs.

docs/proposals/operator-managed-infra.md
```diff
@@ -0,0 +1,116 @@
...
+3. Should we change the nodeSelectors based on whether Karpenter is or isn't installed?
+4. Should we install and manage the Nvidia DaemonSet?
```
brandonjbjelland, line 66: I believe we landed on no for this and similar cluster-level DaemonSets (e.g., nvidia drivers) though cluster configurations like this would be good to verify through the controller. If incompatible configurations or non-existent dependencies are found, the controller should likely emit warning level logs.

I'm still not 100% convinced that SCI installation shouldn't be managed by the core controller. I think that's worth revisiting as we get further.

docs/proposals/operator-managed-infra.md
```diff
@@ -2,73 +2,56 @@
...
+* Simplified install: Allow install to use `gcloud/aws` or `eksctl` to create initial K8s cluster + nodepools
 * Make install onto existing K8s cluster straight forward and remove
   complexity of using Terraform
-* Improved UX, all an end-user needs to do is install our manifests and ensure GCP service account
-  has enough permissions
-* Enterprises generally have very specific requirements
-  about how they create their K8s Clusters and nodepools.
-* Allow Substratus to work in a multi namespace environment.
-  Currently namespace is hardcoded in workload identity settings in the terraform
+* Improved UX, all an end-user needs to do is install our manifests and ensure GCP service account has enough permissions
```
brandonjbjelland, line 19: > ensure GCP service account has enough permissions

I think fleshing out what's meant by this statement will answer what we want here: https://github.com/substratusai/substratus/pull/170#discussion_r1287980291

docs/proposals/operator-managed-infra.md
```diff
@@ -2,73 +2,56 @@
...
+  * read and write a specific bucket
+  * pull and push to a specific Image Registry (GAR/ECR)
+  * (GCP) set IAM policy on the service Account itself to be able to manage workload identity bindings to KSAs. This can be done by assigning the service account to be a ServiceAccountAdmin to itself
+  * (AWS) UpdateAssumeRolePolicy on the AWS Role used by Substratus (required for multi namespace support)
+* REQUIRED: Apply Subsubstratus manifests and configure them to use correct bucket and image registry
+
+### Controller
+Controller will be responsible for the following:
+* Creating the K8s ServiceAccount and related binding by calling `enforceServiceAccount` whenever a Substratus resource gets reconciled
+* (GCP only) example set annotation for Google Service Account AND update IAM policy ton the Service Account so it can use the Google Service Account. For example:
```
brandonjbjelland, line 42:
```suggestion
* (GCP only) example set annotation for Google Service Account AND update IAM policy on the Service Account so it can use the Google Service Account. For example:
```

docs/proposals/operator-managed-infra.md
```diff
@@ -0,0 +1,71 @@
...
+1. (Optional) Create your GKE cluster `gcloud container clusters create` and create nodepools
+
+2. Create service account and assign permissions needed
+
+   ```
+   gcloud iam service-accounts create substratus
+   gcloud iam service-accounts add-iam-policy-binding substratus@my-project.iam.gserviceaccount.com \
+    --role roles/iam.serviceAccountAdmin \
+    --member "serviceAccount:myproject.svc.id.goog[substratus/substratus]"
+   gcloud projects add-iam-policy-binding my-project \
```
nstogner, line 65: Lets avoid project level bindings where we can: https://cloud.google.com/artifact-registry/docs/access-control#grant-repo

docs/proposals/operator-managed-infra.md
```diff
@@ -0,0 +1,71 @@
...
+
+1. (Optional) Create your GKE cluster `gcloud container clusters create` and create nodepools
+
+2. Create service account and assign permissions needed
+
+   ```
+   gcloud iam service-accounts create substratus
+   gcloud iam service-accounts add-iam-policy-binding substratus@my-project.iam.gserviceaccount.com \
+    --role roles/iam.serviceAccountAdmin \
+    --member "serviceAccount:myproject.svc.id.goog[substratus/substratus]"
```
nstogner, line 64: I think `--member` is wrong here, should likely be `--member "serviceAccount:substratus@my-project.iam.gserviceaccount.com"`

nstogner: I am guessing this was to give this SA access to modify its own bindings?

nstogner: PS: We will also need workload identity binding for substratus to start using the SA:

```bash
gcloud iam service-accounts add-iam-policy-binding GSA_NAME@GSA_PROJECT.iam.gserviceaccount.com \
    --role roles/iam.workloadIdentityUser \
    --member "serviceAccount:PROJECT_ID.svc.id.goog[NAMESPACE/KSA_NAME]"
```

docs/proposals/operator-managed-infra.md
```diff
@@ -0,0 +1,71 @@
...
+Controller will be responsible for the following:
+* Creating the K8s ServiceAccount and related binding by calling `enforceServiceAccount` whenever a Substratus resource gets reconciled
+* (GCP only) example set annotation for Google Service Account AND update IAM policy on the Service Account so it can use the Google Service Account. For example:
+  ```
+  gcloud iam service-accounts add-iam-policy-binding substratus@my-project.iam.gserviceaccount.com \
+   --role roles/iam.workloadIdentityUser \
+   --member "serviceAccount:myproject.svc.id.goog[new-namespace/substratus]"
+  ```
+* (AWS Only) Annotate the service account with ARN of IAM role AND (awsmanager) call UpdateAssumeRolePolicy
+* The annotations to service accounts happen inside the controller itself
```
nstogner, line 49: Already stated

nstogner (changes requested):
I think the documented permissions at the end are wrong, other than that, feel free to merge.

docs/proposals/operator-managed-infra.md
samos123: this was the part with a todo that I didn't get to update yet :P I added this at the top "TODO: Update this with actual proposed install steps for GCP"

in hindsight I should have just removed it before asking you all to review

docs/proposals/operator-managed-infra.md
samos123: yep latest version doesn't have this, you weren't supposed to review this part, sorry about that

docs/proposals/operator-managed-infra.md
```diff
@@ -0,0 +1,118 @@
...
+* REQUIRED: Create Google Service Account or AWS IAM Role that has resource level permissions to:
+  * read and write a specific bucket
+  * pull and push to a specific Image Registry (GAR/ECR)
+  * (GCP) set IAM policy on the service Account itself to be able to manage workload identity bindings to KSAs. This can be done by assigning the service account to be a ServiceAccountAdmin to itself
+  * (AWS) UpdateAssumeRolePolicy on the AWS Role used by Substratus (required for multi namespace support)
+* REQUIRED: Apply Substratus manifests and configure them to use correct bucket and image registry
+
+### Controller
+Controller will be responsible for the following:
+* Creating the K8s ServiceAccount (already done today) and related binding by calling `enforceServiceAccount` whenever a Substratus resource gets reconciled
```
brandonjbjelland, line 41: To make the call provider agnostic, I was leaning toward `UpdateIamPrincipal` as the GRPC service we'd implement. The term principal seems to be common terminology across all providers. `Update` is in line with other verbs I've seen in GRPC-land.

```suggestion
* Creating the K8s ServiceAccount (already done today) and related binding by calling `UpdateIamPrincipal` whenever a Substratus resource gets reconciled
```

Variations:

* `UpdateIamPrincipalPolicy`
* `UpdateIdentityPolicy`

docs/proposals/operator-managed-infra.md
```diff
@@ -0,0 +1,118 @@
...
+* (GCP only) example set annotation for Google Service Account (already done today) AND call CSI to update IAM policy on the Service Account so it can use the Google Service Account. For example:
```
brandonjbjelland, line 42:
```suggestion
* (GCP only) example set annotation for Google Service Account (already done today) AND call SCI to update IAM policy on the Service Account so it can use the Google Service Account. For example:
```

docs/proposals/operator-managed-infra.md
```diff
@@ -0,0 +1,118 @@
...
+  ```
+  gcloud iam service-accounts add-iam-policy-binding substratus@my-project.iam.gserviceaccount.com \
+   --role roles/iam.workloadIdentityUser \
+   --member "serviceAccount:myproject.svc.id.goog[new-namespace/substratus]"
+  ```
+* (AWS Only) Annotate the service account with ARN of IAM role AND (awsmanager) use CSI to call UpdateAssumeRolePolicy
```
brandonjbjelland, line 48: This is new as of this afternoon but `sci-${provider}` is better because it's more specific than our current`${provider}-manager`.

```suggestion
* (AWS Only) Annotate the service account with ARN of IAM role AND (sci-aws) use CSI to call UpdateAssumeRolePolicy
```

brandonjbjelland (changes requested):
Just a couple of details to make sure we're aligned.

docs/proposals/operator-managed-infra.md
```diff
@@ -0,0 +1,118 @@
...
+
+Instead of managing all the infrastructure through terraform, the
+operator itself should be responsible for managing the Workload Identity
+bindings between the K8s Service Account and the Cloud Service Account/IAM role.
+
+Doing this, will make it possible to use Substratus resources in multiple
+namespaces without having to use another tool to configure Cloud identity/role mappings.
+
+Out of scope:
+* Managing the bucket, image registry or other any other resources
```
brandonjbjelland, line 12:
```suggestion
* Managing the cluster configuration, node pools, dependent workpads/daemonsets, the bucket, image registry, the `sci-${provider}` principal, and its permissions.
```

docs/proposals/operator-managed-infra.md
samos123: I hadn't thought of this being a generic call but that makes total sense.

samos123: I will rephrase this since we should re-use the existing enforceServiceAccount function but in that function i should call the generic `UpdateIamPrincipal` from `sci-${CLOUD}`

docs/proposals/operator-managed-infra.md
```diff
@@ -0,0 +1,128 @@
...
+  * pull and push to a specific Image Registry (GAR/ECR)
+  * (GCP) set IAM policy on the service Account itself to be able to manage workload identity bindings to KSAs. This can be done by assigning the service account to be a ServiceAccountAdmin to itself
+  * (AWS) UpdateAssumeRolePolicy on the AWS Role used by Substratus (required for multi namespace support)
+* REQUIRED: Apply Substratus manifests and configure them to use correct bucket and image registry
+
+### Controller
+Controller will be responsible for the following:
+* Creating the K8s ServiceAccount (already done today in `enforceServiceAccount`), in addition should now also call `sci-${CLOUD}:BindKSAToIAMPrincipal` to ensure the new KSA has access to the AWS IAM Role/GCP Service Account
+
+* (GCP only) example set annotation for Google Service Account (already done today) AND call `sci-gcp:BindKSAToIAMPrincipal` to update IAM policy on the Service Account so it can use the Google Service Account. For example:
```
brandonjbjelland, line 43: I think the call would just be `sci.BindKSAToIAMPrincipal` across all providers.

Should we separate out the GRPC services for adding and removing bindings? That seems implied in the decision to use `BindKSAToIAMPrincipal`. I was using `UpdateIamPrincipal` because I imagined an `operation: string` argument in the function signature with valid values of either `"add"` or `"remove"` so we'd be able to reuse the same service for both use cases. Let's either decide to:

1. consolidate into a single GRPC service or
2. decide on both add and removal service names (they're both needed) and add a detail on how that would look, what would trigger it, etc.

samos123: I think we should never remove a K8s SA from an IAM principal. So vote for leaving that out of the proposal until we later see a need for it? When should we remove a K8s SA binding from IAM principal? I think it's better to leave those intact until the IAM principal itself is removed.

brandonjbjelland: > When should we remove a K8s SA binding from IAM principal?

If a namespace is deleted

samos123: I'm ore concerned about the risk where an end-user has 2 clusters and in 1 cluster we remove the binding which causes his other cluster to also lose access unexpectedly. Hence I think we should just leave them in-place. Does AWS do cluster level bindings? If so that would be a good argument for implementing both add and remove. In GCP remove simply wouldn't do anything.

brandonjbjelland (dismissed):

docs/proposals/operator-managed-infra.md
```diff
@@ -0,0 +1,128 @@
...
+   --member "serviceAccount:myproject.svc.id.goog[new-namespace/substratus]"
+  ```
+* (AWS Only) Annotate the Kubernetes service account with ARN of IAM role AND (sci-aws) use `sci-aws:BindKSAToIAMPrincipal` to call `UpdateAssumeRolePolicy` to update the role's trust policy.
+* Any API calls made to clouds should go through a cloud manager e.g. `gcpmanager` or `awsmanager`
+
+### Substratus Cloud Interface
+The SCI will have a generic function `BindKSAToIAMPrincipal` that takes the following input:
+```
+namespace: The K8s namespace
+serviceAccount: The name of K8s Service Account
```
brandonjbjelland, line 56:
```suggestion
kubernetesServiceAccount: The name of K8s Service Account
```

docs/proposals/operator-managed-infra.md
```diff
@@ -0,0 +1,128 @@
...
+IAMPrinciple: The GCP service account or IAM role
```
brandonjbjelland, line 57:
```suggestion
IAMPrincipal: The GCP service account or IAM role
```

docs/proposals/operator-managed-infra.md
```diff
@@ -0,0 +1,128 @@
...
+* (AWS Only) Annotate the Kubernetes service account with ARN of IAM role AND (sci-aws) use `sci-aws:BindKSAToIAMPrincipal` to call `UpdateAssumeRolePolicy` to update the role's trust policy.
+* Any API calls made to clouds should go through a cloud manager e.g. `gcpmanager` or `awsmanager`
+
+### Substratus Cloud Interface
+The SCI will have a generic function `BindKSAToIAMPrincipal` that takes the following input:
+```
+Namespace: The K8s namespace
+KubernetesServiceAccount: The name of K8s Service Account
+IAMPrincipal: The GCP service account or IAM role
+Params Map[string]string: Any additional parameters that might be required by a specific cloud provider
```
nstogner, line 58: Prefer to omit Params until we find the need

nstogner (approved):
Looks great, added a single comment, but dont wait to merge

docs/proposals/operator-managed-infra.md
nstogner: Prefer to table cleanup for another day. Lets capture the concerns in an issue.

docs/proposals/operator-managed-infra.md
nstogner: Defer to convo here: https://github.com/substratusai/substratus/pull/177

nstogner: OK with leaving this level of detail out of proposal, or just stating that these things are "first thoughts" in order to get proposal closed.

samos123: I think when possible we should discuss it in the proposal instead of having to significantly change the code PR later. I will add a note saying the actual implementation will likely differ

samos123: agreed with Nick that some of these should be provided as part of configMap or automatically detected at runtime instead of using params. This is such that the controller doesn't have to have different call for each cloud provider

brandonjbjelland: My take: proposals are good but probably shouldn't hammer out fine grain details. Proposals should paint the high-level picture with a rough outline of the implementation plan. Things like naming should often be deferred

samos123: I think the risk with that is renaming in a PR with code sometimes requires full retesting and changing a lot of code. So I rather hash out as many details as possible. The interfaces and names of those interfaces should be discussed in the proposal when possible but not be required.

docs/proposals/operator-managed-infra.md
brandonjbjelland: Captured here: https://github.com/substratusai/substratus/issues/179

PR #177, opened 2023-08-10
adding proto definitions for KSA <=> IAM actions in SCI

brandonjbjelland:
## What is this change?

Adds protobuf definitions and messages for the new service

## Why make this change?

As a part of https://github.com/substratusai/substratus/pull/171


internal/sci/sci.proto
```diff
@@ -1,13 +1,24 @@
 syntax = "proto3";
 
 package sci.v1;
+import "google/protobuf/struct.proto";
 option go_package = "github.com/substratusai/substratus/internal/sci";
 
 service Controller {
   rpc CreateSignedURL(CreateSignedURLRequest) returns (CreateSignedURLResponse) {}
   rpc GetObjectMd5(GetObjectMd5Request) returns (GetObjectMd5Response) {}
+  rpc BindKSAToIAMPrincipal(BindKSAToIAMPrincipalRequest) returns (BindKSAToIAMPrincipalResponse) {} // strictly adds a permission for an IAM principal to a KSA
```
brandonjbjelland, line 10: Tagging @nstogner 

Our casing is a little inconsistent on other services (`Md5` should be `MD5` if `URL` isn't `Url`) so let's take a moment to tidy those up. Some alternatives to the service name here:

* `AddKSAToIAMPrincipalBinding`
* `CreateKSAToIAMPrincipalBinding`
* `PatchIamPrincipalBinding`
* `AssociateKSAWithPrincipal`

Between Sam and I, we're committed to making this service additive-only and not able to specify an operation (e.g. add, remove).

The Request and Response messages fall out of the naming decision here.

internal/sci/sci.proto
```diff
@@ -1,11 +1,25 @@
...
+}
+
+message BindKSAToIAMPrincipalRequest {
+  string namespace = 1;
+  string kubernetesServiceAccount = 2;
+  string principal = 3;
+  google.protobuf.Struct additionalParams = 4;
```
brandonjbjelland, line 17: We started with just `params` I thought maybe `policy` could work. This is somewhat flexible.

TO give a sense of what other cloud providers require:
* Using the `az` command line tool: https://azure.github.io/azure-workload-identity/docs/topics/federated-identity-credential.html#federated-identity-credential-for-a-user-assigned-managed-identity
* Here's the go package and method in AWS: https://docs.aws.amazon.com/sdk-for-go/api/service/iam/#IAM.UpdateAssumeRolePolicy

nstogner (changes requested):

internal/sci/sci.proto
samos123: The issue is that some clouds will require additional information and others do not. However at a minimum all clouds will require namespace, KSA and IAM principal. Any additional details will have to be passed through params.

I think that would make it hard for the controller to know when to submit what information. So in hindsight. I think it makes sense to expand the gRPC call over time and for now add all info needed for AWS and GCP in the actual request.

brandonjbjelland: Updated to reflect what will work for GCP and AWS. AWS actually looks a lot like Azure in this domain so what I've got now should work for our immediate platforms.

internal/sci/sci.proto
brandonjbjelland: `BindIdentity` is short and simple but maybe too much so. What's really happening is establishing a one-way trust relationship between identities. I'm happy to go with `BindIdentity` because this isn't so important - we own the only client of SCI and can change it any time.

internal/sci/sci.proto
```diff
@@ -1,11 +1,25 @@
...
+}
+
+message BindKSAToIAMPrincipalResponse {
+  bool success = 1; // Indicate whether the binding was successful
```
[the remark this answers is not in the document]
brandonjbjelland: done

internal/sci/sci.proto
```diff
@@ -1,11 +1,25 @@
...
 
 service Controller {
   rpc CreateSignedURL(CreateSignedURLRequest) returns (CreateSignedURLResponse) {}
   rpc GetObjectMd5(GetObjectMd5Request) returns (GetObjectMd5Response) {}
+  rpc BindKSAToIAMPrincipal(BindKSAToIAMPrincipalRequest) returns (BindKSAToIAMPrincipalResponse) {} // strictly adds a permission for an IAM principal to a KSA
+}
+
+message BindKSAToIAMPrincipalRequest {
+  string namespace = 1;
+  string kubernetesServiceAccount = 2;
```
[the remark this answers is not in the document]
brandonjbjelland: I don't love `cloud_principal` (`target_principal` is perhaps better) but the rest works for me.

brandonjbjelland:
Related: https://github.com/substratusai/substratus/pull/180

internal/sci/sci.proto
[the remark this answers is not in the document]
samos123: I think cloud_principal or just principal is good too.

samos123 (changes requested):

internal/sci/sci.proto
```diff
@@ -6,8 +6,18 @@ option go_package = "github.com/substratusai/substratus/internal/sci";
...
   rpc CreateSignedURL(CreateSignedURLRequest) returns (CreateSignedURLResponse) {}
   rpc GetObjectMd5(GetObjectMd5Request) returns (GetObjectMd5Response) {}
+  rpc BindIdentity(BindIdentityRequest) returns (BindIdentityResponse) {}
 }
 
+message BindIdentityRequest {
+  string principal = 1; // the target cloud identity (e.g., an AWS IAM role or GSA)
+  string kubernetes_service_account = 2;
+  string kubernetes_namespace = 3;
+  string federated_oidc_provider = 4; // a URL that identifies the trusted OIDC provider for the originating identity (KSA)
```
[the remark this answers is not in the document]
brandonjbjelland: I wasn't so sure you could get it from within the cluster but it looks like it's possible on both EKS and AKS. Good call - Removing.

samos123 (dismissed):

nstogner (dismissed):
Looks good

Issue #176, opened 2023-08-10
`make dev-down-aws` is blocked when deleting clusters having instances created by karpenter


brandonjbjelland:
I'm seeing the following error as eksctl tries to tear down the cloudformation stack:
```
ClusterSharedNodeSecurityGroup | DELETE_FAILED | resource sg-05d7bc8cce51e6371 has a dependent object (Service: AmazonEC2; Status Code: 400; Error Code: DependencyViolation; Request ID: adae33a8-d2ba-482d-8833-76365f2ccc55; Proxy: null)
```
I gather the node in question is using a security group that the stack creates. After deleting the GPU instance, the stack deletion worked as expected. The instance itself was named in such a way that we could identify `karpenter` as the owner but I didn't check the tags to see if there's any other metadata. If there isn't a tag that gives us this information, we should modify the Provisioner to add a tag that identifies the `cluster` so we can delete all instances tagged and named in such a way prior to running `make dev-down-aws` or similar.

brandonjbjelland:
Resolved

PR #181, opened 2023-08-10
launch karpenter instances into private subnets, terminates karpenter instances in `aws-down.sh`

brandonjbjelland:
## What is this change?

This change does 2 things:

1. adds configuration to deploy karpenter nodes into private subnets
2. destroying karpenter instances by tag reference prior to `eksctl delete`

## Why make this change?

1. using public subnets gives external IPs to nodes - this is an insecure practice and unnecessary when instances have a NAT gateway to egress through
2. We need to terminate all instances prior to CFN stack deletion or it hangs since the security group is in use. https://github.com/substratusai/substratus/issues/176


samos123 (approved):

nstogner (approved):
Looks good!

PR #182, opened 2023-08-10
Fix the flaky `TestModelTrainerFromGit`and `TestModelLoaderFromGit` tests

brandonjbjelland:
## What is this change?

This change adds some retry logic to a test that seems to fail in github actions regularly (though never locally). This may take a few iterations since I can't repro on my workstation.

## Why make this change?

We are regularly seeing this test fail because the models don't reconcile quickly enough. I dont think this is a problem we've seen happen in the wild so I think this is more of a test issue than an actual problem with the reconciler. The error:

```
the object has been modified; please apply your changes to the latest version and try again
```

Recent failures:
https://github.com/substratusai/substratus/actions/runs/5826988660/job/15802093114
https://github.com/substratusai/substratus/actions/runs/5827031817/job/15802216794

### DIFF
--- a/internal/controller/model_controller_test.go
+++ b/internal/controller/model_controller_test.go
@@ -20,7 +20,7 @@ import (
 
 func RetryOnConflictError(operation func() error) error {
 	backoff := retry.DefaultBackoff
-	backoff.Steps = 5
+	backoff.Steps = 10
 
 	return retry.OnError(backoff, func(err error) bool {
 		return errors.IsConflict(err)
@@ -47,7 +47,10 @@ func TestModelLoaderFromGit(t *testing.T) {
 			},
 		},
 	}
-	require.NoError(t, k8sClient.Create(ctx, model), "create a model that references a git repository")
+	err := RetryOnConflictError(func() error {
+		return k8sClient.Create(ctx, model)
+	})
+	require.NoError(t, err, "create a model that references a git repository")
 
 	testContainerBuild(t, model, "Model")
 	testParamsConfigMap(t, model, "Model", `{ "s": "something-model", "x": 456 }`)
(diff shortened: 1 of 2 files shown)


internal/controller/model_controller_test.go
```diff
@@ -9,13 +9,24 @@ import (
...
 )
 
+func RetryOnConflictError(operation func() error) error {
+	backoff := retry.DefaultBackoff
+	backoff.Steps = 5
+
+	return retry.OnError(backoff, func(err error) bool {
+		return errors.IsConflict(err)
+	}, operation)
+}
```
brandonjbjelland, line 28: We don't have a pattern established just yet of where to put common test utilities like this. I could see a `testutil` package becoming useful but we can probably hold off until we see it (or similar functions) reused.

brandonjbjelland:
On average this action looks like it fails about 1/3 of the time. Almost all of those failures in recent times cite this test failing. I've just run the `Integration tests/unit-tests` action 5 times in a row without failure.

samos123 (dismissed):

nstogner (commented):
Suspicious that a create would ever need a retry on conflict... I cant come up with a scenario in which that would be the case.

NOTE: `the object has been modified; please apply your changes to the latest version and try again` is a normal error that the reconcilers throw... unlikely to be the root cause here. I am guessing there is something else that is the cause. Will look into the test logs more soon.

brandonjbjelland:
Yeah I wasn't convinced by the new fail rate to call this done. I'll look a bit further here.

brandonjbjelland:
Put our heads together and "solved" this here: https://github.com/substratusai/substratus/pull/184

PR #185, opened 2023-08-11
GCP: Updated install flow and managed infra

samos123:
* Changes the Service Account that the controller-manager and gcpmanager use to `substratus
* Removed all terraform except parts related to cluster and nodepool creation
* Updated gcp-up.sh to create bucket, gar, GSA and required permissions
* Add skaffold support gcp for both the controller-manager and csi

The GCP branch is working very good so far. Here is what I tested:
* New installation flow brings up cluster and script with gcloud commands adds required permissions
make dev-run-gcp and verifying workload identity things got installed
* in-cluster deployment by using make dev-skaffold-gcp and doing the following in a new namespace:
  * Load a model
  * Finetune a model
  * Load a dataset
  *  Serve a model
  * Run a notebook using upload
 
Issue has been resolved by retrying and sleeping for 5 seconds during initial startup of sci-gcp. There was a long delay after annotating the K8s SA for it to be able to utilize the GSA. As a result the sci-gcp container will restart many times until GKE workload identity finally starts working. You would originally see the following error message:
```
ERROR   setup   failed to validate server       {"error": "Post \"https://iam.googleapis.com/v1/projects/sam-XX/serviceAccounts/substratus@sam-XX.iam.gserviceaccount.com:getIamPolicy?alt=json&prettyPrint=false\": compute: Received 403 `Unable to generate access token; IAM returned 403 Forbidden: Permission 'iam.serviceAccounts.getAccessToken' denied on resource (or it may not exist).\nThis error could be caused by a missing IAM policy binding on the target IAM service account.\nFor more information, refer to the Workload Identity documentation:\n\thttps://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#authenticating_to\n\n`"}
```

Issue: Kaniko builder gets stuck when creating a new namespace
```
error checking push permissions -- make sure you entered the correct tag name, and that you are authenticated correctly, and try again: checking push permission for "us-central1-docker.pkg.de...
```
Solved by running an initContainer that validates GKE metadata is ready for the pod: https://cloud.google.com/kubernetes-engine/docs/troubleshooting/troubleshooting-security#workload-identity


config/default/manager_patch.yaml
```diff
@@ -51,7 +51,7 @@ spec:
         - name: manager
           envFrom:
             - configMapRef:
-                name: system
```
nstogner, line 54: The name of the ConfigMap should likely still be "system" within the "substratus" namespace.

config/gcpmanager/gcp-manager.yaml
```diff
@@ -19,7 +13,7 @@ spec:
       labels:
         app: gcp-manager
     spec:
-      serviceAccountName: gcp-manager
+      serviceAccountName: substratus
```
nstogner, line 16: Prefer to use "sci" for this serviceAccountName. This should not be the same as the controller manager. Keeping the controller manager free of cloud privileges is better from a security perspective.

nstogner (changes requested):

config/default/manager_patch.yaml
samos123: fixed

config/gcpmanager/gcp-manager.yaml
samos123: fixed

internal/sci/gcp/manager_test.go
```diff
@@ -0,0 +1,104 @@
...
+	err = server.AutoConfigure(mc)
+	if err != nil {
+		t.Errorf("Error running AutoConfigure %v", err)
+	}
+
+	expectedMember := fmt.Sprintf("serviceAccount:%s.svc.id.goog[integration-test/integration-test]", server.ProjectID)
+	resourceID := fmt.Sprintf("projects/%s/serviceAccounts/%s", server.ProjectID, server.SaEmail)
+	// Get current policy and remove bindings left behind from previous tests
+	policy, err := server.Clients.IAM.Projects.ServiceAccounts.GetIamPolicy(resourceID).Context(ctx).Do()
+	if err != nil {
```
nstogner, line 47: Prefer using `require` pkg for asserting ... `require.NoError` ... See other tests.

internal/sci/gcp/manager_test.go
```diff
@@ -0,0 +1,104 @@
...
+		t.Errorf("error getting IAM policy of service account: %v", err)
+	}
+	logIAMPolicyBindings(t, policy.Bindings, "policy bindings before BindIdentity call")
+
+	policy.Bindings = cleanUpBinding(t, policy.Bindings, expectedMember)
```
nstogner, line 52: `cleanUpBinding` was ambiguous to me, prefer `removeBindingMember`

internal/sci/gcp/manager_test.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	policy, err = server.Clients.IAM.Projects.ServiceAccounts.GetIamPolicy(resourceID).Context(ctx).Do()
+	if err != nil {
+		t.Errorf("error getting IAM policy of service account: %v", err)
+	}
+	logIAMPolicyBindings(t, policy.Bindings, "policy bindings after BindIdentity")
+	bindingWasSet := false
+	for _, binding := range policy.Bindings {
+		if slices.Contains(binding.Members, expectedMember) {
```
nstogner, line 76: See `require.Contains`

nstogner (changes requested):
Can you remove `config/sci-kind` and `config/install-kind`? These will cause merge conflicts later.

internal/sci/gcp/manager_test.go
samos123: fixed, that is indeed much cleaner. Thanks!

internal/sci/gcp/manager_test.go
samos123: fixed

internal/sci/gcp/manager_test.go
samos123: I don't think that would work since it's happening in a for loop? Show me the light. I might be missing something obvious.

config/install-gcp/manager_patch.yaml
brandonjbjelland, line 1: I don't understand why this is a patch and not just an unpatched manifest you'd use always

install/scripts/gcp-up.sh
```diff
@@ -39,6 +40,53 @@ cluster_region=$(terraform output --raw cluster_region)
 
 cd -
 
+# Create a bucket for substratus models and datasets
+export ARTIFACTS_BUCKET="gs://${PROJECT}-substratus-artifacts"
+if ! gcloud storage buckets describe "${ARTIFACTS_BUCKET}" -q >/dev/null; then
+  gcloud storage buckets create --project ${PROJECT} "${ARTIFACTS_BUCKET}" --location ${cluster_region}
```
brandonjbjelland, line 46: does this succeed if the bucket already exists?

install/scripts/gcp-up.sh
brandonjbjelland, line 1: I expected to see a companion `gcp-down.sh`

internal/sci/gcp/manager_test.go
brandonjbjelland: I think it's adding an import:
```
import (
	"github.com/stretchr/testify/require"
)
```

And using its function:

```suggestion
		require.Contains(t, binding.Members, expectedMember)
		bindingWasSet = true
```

brandonjbjelland (dismissed):
If this is fully working, awesome work and ship it when you're ready.

internal/sci/gcp/manager_test.go
samos123: require.Contains would throw a failure the moment it's not contained so that wouldn't work?

install/scripts/gcp-up.sh
samos123: oh right I hadn't gotten to it. Good catch!

config/install-gcp/manager_patch.yaml
samos123: I think because possibly not everywhere the rbac proxy is supported

cmd/controllermanager/main.go
```diff
@@ -48,7 +48,7 @@ func main() {
 	var sciAddr string
 	flag.StringVar(&configDumpPath, "config-dump-path", "", "The filepath to dump the running config to.")
 	// TODO: Change SCI Service name to be cloud-agnostic.
-	flag.StringVar(&sciAddr, "sci-address", "gcp-manager.substratus.svc.cluster.local:10080", "The address of the Substratus Cloud Interface server.")
+	flag.StringVar(&sciAddr, "sci-address", "sci.substratus.svc.cluster.local:10080", "The address of the Substratus Cloud Interface server.")
```
nstogner, line 51: It looks like this is already on main? https://github.com/substratusai/substratus/blob/75467f6f643fa6e9b86cb4643256e21f43e1c40c/cmd/controllermanager/main.go#L51

internal/sci/gcp/manager.go
```diff
@@ -0,0 +1,188 @@
...
+	Metadata             *metadata.Client
+	Storage              *storage.Client
+	Http                 *http.Client
+}
+
+func NewServer() (*Server, error) {
+	s := &Server{}
+	ctx := context.Background()
+	if err := envconfig.Process(ctx, s); err != nil {
+		return nil, fmt.Errorf("environment: %w", err)
```
nstogner, line 44: Prefer to see this in main()

internal/sci/gcp/manager.go
```diff
@@ -0,0 +1,188 @@
...
+
+type Clients struct {
+	IAMCredentialsClient *credentials.IamCredentialsClient
+	IAM                  *iam.Service
+	Metadata             *metadata.Client
+	Storage              *storage.Client
+	Http                 *http.Client
+}
+
+func NewServer() (*Server, error) {
```
nstogner, line 40: NewServer not needed, clearer in main()

internal/sci/gcp/manager.go
```diff
@@ -0,0 +1,188 @@
...
+func (s *Server) CreateSignedURL(ctx context.Context, req *sci.CreateSignedURLRequest) (*sci.CreateSignedURLResponse, error) {
+	log := log.FromContext(ctx)
+	log.Info("creating signed URL", "bucket", req.BucketName, "object", req.ObjectName)
+
+	bucketName, objectName, checksum := req.GetBucketName(),
+		req.GetObjectName(),
+		req.GetMd5Checksum()
+	bucket := s.Clients.Storage.Bucket(bucketName)
+	obj := bucket.Object(objectName)
+	if _, err := obj.Attrs(ctx); err != nil && err != storage.ErrObjectNotExist {
```
nstogner, line 59: What is this used for?

skaffold.gcp.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: skaffold/v4beta6
+kind: Config
+manifests:
+  kustomize:
+    paths:
+      - ./config/skaffold-gcp
+build:
+  artifacts:
+    - image: controller-manager
```
nstogner, line 9: Where is skaffold going to push these images to?

internal/sci/gcp/manager.go
samos123: I think Brandon did this to check for errors by seeing if it worked to get attrs? @brandonjbjelland might be able to explain. This is coming over from original gcpmanager code.

cmd/controllermanager/main.go
samos123: yes gone after rebase

Makefile
```diff
@@ -286,15 +274,19 @@ uninstall-crds: manifests kustomize ## Uninstall CRDs from the K8s cluster speci
 installation-scripts:
 	perl -pi -e "s/version=.*/version=$(VERSION)/g" install/scripts/install-kubectl-plugins.sh
 
-.PHONY: installation-manifests
-installation-manifests: manifests kustomize
+.PHONY: installation-manifests-gcp
+installation-manifests-gcp: manifests kustomize
+	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
```
nstogner, line 279: This is duplicated in each one of these makefile targets. Also by removing the original target we have broken the `prepare-release` target. Prefer to keep one target here to avoid Makefile sprawl.

internal/controller/utils.go
```diff
@@ -74,3 +78,22 @@ func paramsToEnv(params map[string]intstr.IntOrString) []corev1.EnvVar {
 	}
 	return envs
 }
+
+func CreateKubernetesClient() (*kubernetes.Clientset, error) {
```
nstogner, line 82: Client setup should happen in `main()` and not be hidden inside of pkg helper functions. As a reader, I would like to be able to open `main.go` and see all configuration settings then trace those settings through to all dependency clients as they are constructed, then see those clients passed into Reconciler instances, etc. We have diverged from this with some of the AutoConfigure stuff we are doing in the `cloud` pkg, but we should not continue expanding in that direction where its not necessary. I will likely refactor some of the cloud stuff to get use closer to this described pattern later.

Example: https://github.com/ardanlabs/service/blob/master/app/services/sales-api/main.go

internal/controller/utils.go
```diff
@@ -74,3 +78,22 @@ func paramsToEnv(params map[string]intstr.IntOrString) []corev1.EnvVar {
...
+	// Try in-cluster config
+	config, err := rest.InClusterConfig()
```
nstogner, line 84: No need to do this, grab this from `mgr.GetConfig()` inside of `main()`.

internal/sci/gcp/manager.go
```diff
@@ -0,0 +1,204 @@
...
+	SaEmail   string
+	ProjectID string `env:"PROJECT_ID"`
+}
+
+type Clients struct {
+	IAMCredentialsClient *credentials.IamCredentialsClient
+	IAM                  *iam.Service
+	Metadata             *metadata.Client
+	Storage              *storage.Client
+	Http                 *http.Client
```
nstogner, line 37: `HTTP`

internal/sci/gcp/manager_test.go
nstogner: Ahh, I see, disregard

internal/sci/gcp/manager_test.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/substratusai/substratus/internal/sci"
+	"github.com/substratusai/substratus/internal/sci/gcp"
+	"google.golang.org/api/iam/v1"
+
+	// Can be changed to slices once we go to 1.21
+	"golang.org/x/exp/slices"
+)
+
+func TestServer(t *testing.T) {
```
nstogner, line 20: How do I call this? Will this fail if the SA was not pre-created and I run `make test`?

nstogner (changes requested):

internal/controller/utils.go
samos123: fixed, big improvement to do it this way. Thanks for catching these  kind of improvements.

internal/sci/gcp/manager.go
samos123: fixed

internal/sci/gcp/manager_test.go
samos123: you call it by running `make test` and ensuring you've set `export GOOGLE_APPLICATION_CREDENTIALS=$PWD/secrets/substratus-sa.json`

It won't fail if you have not set `GOOGLE_APPLICATION_CREDENTIALS`, it will  simply skip the test. So as soon as our CI starts loading `GOOGLE_APPLICATION_CREDENTIALS` this tests will automatically start running.

internal/sci/gcp/manager.go
brandonjbjelland: My thought process was that it was a good idea to check what's at the current path and what we should expect is that no object exists. If we encounter some other error (permissions or whatever), I think it's a good indication that we wouldn't be able to create the signed URL or have it be used without error by the uploader.

At some point I think we should change how we do uploading to include the MD5checksum in the path, and verify that an object doesn't exist here prior to creating an upload URL (if it does, it means we'd be doing a re-upload and we can likely short circuit the upload process). As it currently stands, if I do two notebook uploads in rapid succession (with some change in between) or if two people are working on the same notebook and performing uploads, I think the state is indeterminate.

internal/controller/utils.go
samos123: this whole function was removed. So fixed.

config/install-gcp/manager_patch.yaml
brandonjbjelland: I see. thanks for explaining 👍

internal/sci/gcp/manager.go
samos123: I'm going to ignore this since changing the logic of signedURL wasn't in scope of this PR. Feel free to open a new issue and our PM can consider prioritization of it.

brandonjbjelland: @nstogner ☝️  thoughts?

.github/workflows/docker-build-push.yaml
```diff
@@ -66,7 +66,7 @@ jobs:
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta-installer.outputs.tags }}
           labels: ${{ steps.meta-installer.outputs.labels }}
-  gcp-manager:
+  csi-gcp:
```
brandonjbjelland, line 69:
```suggestion
  sci-gcp:
```

samos123: I did a search and I used csi a lot, I pushed over your commit and will tackle this by doing find and going through all occurences

PR #186, opened 2023-08-11
WIP: first pass at awsmanager

brandonjbjelland:


internal/sci/aws/awssci.go
```diff
@@ -0,0 +1,268 @@
+// Package aws provides an AWS implementation of the Substratus Cloud Interface (SCI)
+package awssci
```
samos123, line 2: I am just calling it `gcp`  instead of `gcpcsi`

samos123: just calling out a difference, not sure what's recommended here ha

brandonjbjelland: Calling it aws would conflict with my imports. I thought this was the right distinction but maybe aliasing the imports is preferred

nstogner: Best to match dir and alias on import to keep with conventions, gophers like their conventions

brandonjbjelland: can do

internal/gcpmanager/manager.go
```diff
@@ -1,4 +1,4 @@
-// Package gcp provides a GCP implementation of the Substratus Cloud Interface (SCI).
+// Package gcpmanager provides a GCP implementation of the Substratus Cloud Interface (SCI).
```
samos123, line 1: Please remove any changes here. I incorporated them in my PR

internal/sci/aws/awssci.go
samos123: Ok +1 to match dit and use alias for import

internal/sci/aws/server.go
```diff
@@ -0,0 +1,285 @@
...
+	OIDCProviderARN string
+	Clients
+}
+
+type Clients struct {
+	S3Client  *s3.S3
+	IamClient *iam.IAM
+}
+
+func NewAWSServer() (*Server, error) {
```
nstogner, line 38: Convention is to construct clients in main() and pass in to make it very clear what a program is depending on. Typically best to avoid the combo of BOTH configuring AND constructing clients from within pkg functions.

internal/sci/aws/server.go
```diff
@@ -0,0 +1,285 @@
...
+
+	return &Server{
+		Clients:         *c,
+		OIDCProviderURL: OIDCProviderURL,
+		OIDCProviderARN: OIDCProviderARN,
+	}, nil
+}
+
+func (s *Server) GetObjectMd5(ctx context.Context, req *sci.GetObjectMd5Request) (*sci.GetObjectMd5Response, error) {
+	headResult, err := s.getObjectMetadata(s.Clients.S3Client, req.BucketName, req.ObjectName)
```
nstogner, line 77: Prefer to avoid this helper function and move its contents to here. Currently the helper function maps different return conditions into different combo's of nil return values. It would be clearer to do all of that here. There appear to be no other callers of `getObjectMetadata` so it has no re-use benefit.

internal/sci/aws/server.go
```diff
@@ -0,0 +1,285 @@
...
+	// Decode the current trust policy
+	var existingTrustPolicy map[string]interface{}
+	if err := json.Unmarshal([]byte(*getRoleOutput.Role.AssumeRolePolicyDocument), &existingTrustPolicy); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal trust policy: %v", err)
+	}
+
+	subValue := fmt.Sprintf("system:serviceaccount:%s:%s", req.KubernetesNamespace, req.KubernetesServiceAccount)
+
+	// Construct the new trust relationship
+	newTrustRelationship := map[string]interface{}{
```
nstogner, line 166: Prefer to move this to where it is used.

internal/sci/aws/server.go
```diff
@@ -0,0 +1,285 @@
...
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return "", fmt.Errorf("failed to create clientset: %w", err)
+	}
+
+	configMapName := "substratus-global"
+	configMapNamespace := "substratus"
+	configMapKey := "cluster_id"
+
+	configMap, err := clientset.CoreV1().ConfigMaps(configMapNamespace).Get(context.TODO(), configMapName, metav1.GetOptions{})
```
nstogner, line 250: Prefer mounting ConfigMaps into Pod via env or file over grabbing the ConfigMap. SCI should not need to be assigned Kubernetes RBAC.

internal/sci/aws/server.go
```diff
@@ -0,0 +1,285 @@
...
+		return nil, fmt.Errorf("failed to get region: %w", err)
+	}
+
+	stsSvc := sts.New(sess)
+	accountId, err := getAccountID(stsSvc)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get account ID: %w", err)
+	}
+
+	OIDCProviderURL := fmt.Sprintf("oidc.eks.%s.amazonaws.com/id/%s", region, clusterID)
```
nstogner, line 61: Prefer lowercase-first variables here. Uppercase first-variables signify exporting and are unconventional to see inside of functions.

samos123:
I think this should have integration tests that demonstrate sci-aws works as expected. You will probably have to ensure there are 2 ways to set things like cluster id:
1. Use the discovery when on AWS
2. Ability to use environment variables or directly set them on the struct itself

 I also would prefer to keep this PR/branch open until this works end-to-end.

internal/sci/aws/server.go
samos123: Rename `NewServer` because the struct name is `Server`

internal/sci/aws/server.go
samos123: I think we should ensure the system configmap gets mounted as env variables to sci container as well

samos123: I did this in my PR just now, you  could cherry-pick this later if my PR isn't merged yet:
https://github.com/substratusai/substratus/pull/185/commits/adeb7832b4c0f28baf67ec10f4ec8d418f40a452

internal/gcpmanager/manager.go
brandonjbjelland: Can do.

internal/sci/aws/server.go
```diff
@@ -0,0 +1,285 @@
...
+	}, nil
+}
+
+func (s *Server) CreateSignedURL(ctx context.Context, req *sci.CreateSignedURLRequest) (*sci.CreateSignedURLResponse, error) {
+	bucketName, objectName, checksum := req.GetBucketName(),
+		req.GetObjectName(),
+		req.GetMd5Checksum()
+
+	// verify that the object doesn't exist
+	if _, err := s.getObjectMetadata(s.Clients.S3Client, bucketName, objectName); err != nil {
```
brandonjbjelland, line 100: Here's the other caller 😄

internal/sci/aws/server.go
brandonjbjelland: Here's the other caller: https://github.com/substratusai/substratus/pull/186/files#r1292987508

Does that change anything?

internal/sci/aws/server.go
brandonjbjelland: I see, yeah that makes good sense. I'll go that path

internal/sci/aws/server.go
brandonjbjelland: Done

internal/sci/aws/server.go
brandonjbjelland: can do!

internal/sci/aws/server.go
nstogner: `// verify that the object doesn't exist` -> Not following how this call is doing this. Btw, Is that desired behavior? How would a notebook re-upload to re-build?

brandonjbjelland: My confusion here. I think at some point we should do checks to be sure we aren't overwriting objects but not until we have good guarantees on path uniqueness (eg md5checksum as a part of the object key path)

internal/sci/aws/server.go
brandonjbjelland: With the removal of the object existence check, the helper went away. 🤘

internal/sci/aws/server.go
brandonjbjelland: Using env vars now. Its up to the caller to set them. Can do that in follow up or in this PR if we want controller cloud config captured here

internal/sci/aws/awssci.go
brandonjbjelland: done

internal/sci/aws/server.go
```diff
@@ -0,0 +1,218 @@
...
+				alreadyExists = true
+				break
+			}
+		}
+	}
+
+	// Construct the new trust relationship
+	newTrustRelationship := map[string]interface{}{
+		"Effect": "Allow",
+		"Principal": map[string]interface{}{
```
samos123, line 147: could this be `map[string]string` ? I think so because `OIDCProviderARN` is of type `string`

internal/sci/aws/server_test.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	t.Logf("Debug: Current Trust Policy: %s", *getRoleOutput.Role.AssumeRolePolicyDocument)
+
+	resp, err := server.BindIdentity(context.TODO(), &sci.BindIdentityRequest{
+		Principal:                roleName,
+		KubernetesNamespace:      "test-namespace",
+		KubernetesServiceAccount: "test-serviceaccount",
+	})
+	assert.NoError(t, err)
+	assert.NotNil(t, resp)
```
samos123, line 146: I saw how painful it is to decode the string into json so if you feel it's too much effort, feel free to ignore this comment.

There should be an assertion that verifies that after BindIdentity, the GetRole returns an AssumeRolePolicyDocument that includes `system:serviceaccount:test-namespace:test-serviceaccount` sub somewhere in the assumerole policy.

samos123 (commented):
Left a few comments but none of them are blockers

internal/sci/aws/server.go
```diff
@@ -0,0 +1,218 @@
...
+	bucketName, objectName, checksum := req.GetBucketName(),
+		req.GetObjectName(),
+		req.GetMd5Checksum()
+
+	// ensure the object is accessible
+	input := &s3.HeadObjectInput{
+		Bucket: awsSdk.String(bucketName),
+		Key:    awsSdk.String(objectName),
+	}
+	_, err := s.Clients.S3Client.HeadObject(input)
```
nstogner, line 71: What does this check tell us after we get past it?

internal/sci/aws/server_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+	bucket := "substratus-test-bucket-" + randomString(8, charset)
+	object := "test-object"
+
+	_, err = server.Clients.S3Client.CreateBucket(&s3.CreateBucketInput{
+		Bucket: &bucket,
+	})
+	assert.NoError(t, err)
+
+	defer server.Clients.S3Client.DeleteBucket(&s3.DeleteBucketInput{Bucket: &bucket})
```
nstogner, line 60: Improvement: wrap in anonymous function which logs an error if it occurs

internal/sci/aws/server_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+	resp, err := server.GetObjectMd5(context.TODO(), &sci.GetObjectMd5Request{
+		BucketName: bucket,
+		ObjectName: object,
+	})
+	assert.NoError(t, err)
+	assert.NotNil(t, resp)
+	if resp != nil {
+		assert.NotEmpty(t, resp.Md5Checksum)
+	}
```
nstogner, line 78: TODO: Send http request to upload object.

internal/sci/aws/server_test.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func TestBindIdentity(t *testing.T) {
```
nstogner, line 81: Not necessary, but if interested: see consolidation of setup / cleanup using: https://pkg.go.dev/testing#hdr-Main

internal/sci/aws/server_test.go
```diff
@@ -0,0 +1,147 @@
...
+	envClusterName := os.Getenv("CLUSTER_NAME")
+	if envClusterName == "" {
+		t.Skip("Skipping TestBindIdentity because CLUSTER_NAME is not set")
+	}
+
+	sess, err := session.NewSession()
+	assert.NoError(t, err)
+	iamClient := iam.New(sess)
+
+	oidcProviderURL := "oidc.eks.us-west-2.amazonaws.com/id/C2A3CBF5FF8C55D72C8843756CD44444"
```
nstogner, line 95: Where does this URL come from?

internal/sci/aws/server_test.go
nstogner: Lazy way to accomplish: check that it contains a string in the stringified document

internal/sci/aws/server.go
```diff
@@ -0,0 +1,218 @@
...
+	if err != nil {
+		if awsErr, ok := err.(awserr.Error); ok {
+			if awsErr.Code() != s3.ErrCodeNoSuchKey {
+				return nil, fmt.Errorf("failed to head object: %w", err)
+			}
+		} else {
+			// It's an error of a different type, not an awserr.Error
+			return nil, fmt.Errorf("failed to head object: %w", err)
+		}
+	}
```
brandonjbjelland, line 81: Like in the gcs sci, I added this to catch errors related to accessing this key path.

Some S3 errors we might catch:

Permission issues (maybe the provided credentials don't have the required permission).
Network issues (maybe there's an issue connecting to the S3 service).
Bucket issues (unlikely but potentially the bucket doesn't exist).
AWS SDK or client-related issues.

Non-AWS Errors:
Errors not related to the AWS SDK. Potentially lower-level issues.

internal/sci/aws/server.go
brandonjbjelland: Like in the gcs sci, I added this to catch errors related to accessing this key path.

Some S3 errors we might catch:

Permission issues (maybe the provided credentials don't have the required permission). Network issues (maybe there's an issue connecting to the S3 service). Bucket issues (unlikely but potentially the bucket doesn't exist). AWS SDK or client-related issues.

Non-AWS Errors: Errors not related to the AWS SDK. Potentially lower-level issues

internal/sci/aws/server.go
brandonjbjelland: Principal is actually weirdly flexible as a string or an array of strings. I think the empty interface is the right way to handle that. This is valid:
```
"Principal": {
    "AWS": ["arn:aws:iam::123456789012:user/Bob", "arn:aws:iam::987654321098:root"]
}
Also valid:
```
"Principal": {
    "AWS": "arn:aws:iam::123456789012:root",
    "Service": "ec2.amazonaws.com",
    "Federated": "cognito-identity.amazonaws.com"
}
```
```

internal/sci/aws/server.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func GetAccountID(stsSvc *sts.STS) (string, error) {
+	result, err := stsSvc.GetCallerIdentity(&sts.GetCallerIdentityInput{})
+	if err == nil {
+		return *result.Account, nil
+	}
+
+	// Fall back to the environment variable if the sts:GetCallerIdentity call fails
+	envAccountID := os.Getenv("AWS_ACCOUNT_ID")
```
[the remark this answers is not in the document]
brandonjbjelland: Huh - I didn't realize that lib existed. It's not in the stdlib but looks like it handles more complex env var use cases. I suppose since we're consuming 2 vars in this SCI, it may be worth it.

internal/sci/aws/server.go
```diff
@@ -0,0 +1,218 @@
...
+	if clusterID != "" {
+		return clusterID, nil
+	}
+
+	return clusterID, nil
+}
+
+func GetOidcProviderUrl(sess *session.Session, clusterName string) (string, error) {
+	svc := eks.New(sess)
+	input := &eks.DescribeClusterInput{
```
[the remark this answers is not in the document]
brandonjbjelland: Great point! Turns out AWS IAM can do this:
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": "eks:DescribeCluster",
            "Resource": "arn:aws:eks:region:account-id:cluster/your-cluster-name"
        }
    ]
}
```
I don't even think I had a dedicated permission for describe cluster so I'll work through that now. Good reminder! :)

internal/sci/aws/server.go
```diff
@@ -0,0 +1,218 @@
...
+type Server struct {
+	sci.UnimplementedControllerServer
+	OIDCProviderURL string
+	OIDCProviderARN string
+	Clients
+}
+
+type Clients struct {
+	S3Client  *s3.S3
+	IamClient *iam.IAM
```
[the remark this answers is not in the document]
brandonjbjelland: Done!

internal/sci/aws/server_test.go
brandonjbjelland: :cool:

brandonjbjelland: :cool:

internal/sci/aws/server_test.go
brandonjbjelland: This is an example of a URL that we'd see returned from GetOidcProviderUrl

internal/sci/aws/server_test.go
brandonjbjelland: Going to circle back on this one

internal/sci/aws/server_test.go
brandonjbjelland: Done

samos123 (approved):

internal/sci/aws/server.go
samos123: I would prefer to catch permission issues in Validate function that only runs once on startup. That's what I ended up having to do for sci-gcp.

nstogner: Lets remove this check. We are adding network calls for every request and it seems counter-intuitive to me as a reader to run a check that looks up an object that will not exist during happy path. If the bucket does not exist, I would expect the following call to generate a signed url to fail. If there is a networking issue it will be log eventually, perhaps in the next call to create a signed url, perhaps if a call to get the md5. If the SCI is lacking permissions that too will surface when fetching the md5.

nstogner (changes requested):

internal/sci/aws/server_test.go
brandonjbjelland: Wrong test to add it but yes. I added that test and a request.

PR #188, opened 2023-08-11
use substratus namespace in k8s config

samos123:
I verified that the following command results in no changes:
```
make installation-manifests
```

So I think it should cause no issues


brandonjbjelland (approved):
12 results matches my local search. lgtm

PR #189, opened 2023-08-11
reconcile service account call SCI

samos123:
* Call sci.BindIdentity to allow a K8s Service account to impersonate an identity
* Created a FakeSCIControllerClient to be able to run tests easily

### DIFF
--- a/cmd/controllermanager/main.go
+++ b/cmd/controllermanager/main.go
@@ -114,12 +114,13 @@ func main() {
 	}
 	defer conn.Close()
 	// Create a client using the connection
-	gc := sci.NewControllerClient(conn)
+	sciClient := sci.NewControllerClient(conn)
 
 	if err = (&controller.ModelReconciler{
 		Client: mgr.GetClient(),
 		Scheme: mgr.GetScheme(),
 		Cloud:  cld,
+		SCI:    sciClient,
 		ParamsReconciler: &controller.ParamsReconciler{
 			Scheme: mgr.GetScheme(),
 			Client: mgr.GetClient(),
@@ -132,7 +133,7 @@ func main() {
 		Scheme:    mgr.GetScheme(),
 		Client:    mgr.GetClient(),
 		Cloud:     cld,
-		SCI:       gc,
+		SCI:       sciClient,
 		NewObject: func() controller.BuildableObject { return &apiv1.Model{} },
 		Kind:      "Model",
 	}).SetupWithManager(mgr); err != nil {
@@ -143,6 +144,7 @@ func main() {
 		Client: mgr.GetClient(),
 		Scheme: mgr.GetScheme(),
 		Cloud:  cld,
+		SCI:    sciClient,
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "Server")
 		os.Exit(1)
@@ -151,7 +153,7 @@ func main() {
 		Scheme:    mgr.GetScheme(),
 		Client:    mgr.GetClient(),
 		Cloud:     cld,
-		SCI:       gc,
+		SCI:       sciClient,
 		NewObject: func() controller.BuildableObject { return &apiv1.Server{} },
 		Kind:      "Server",
 	}).SetupWithManager(mgr); err != nil {
@@ -162,6 +164,7 @@ func main() {
 		Client: mgr.GetClient(),
 		Scheme: mgr.GetScheme(),
 		Cloud:  cld,
+		SCI:    sciClient,
 		ParamsReconciler: &controller.ParamsReconciler{
 			Scheme: mgr.GetScheme(),
 			Client: mgr.GetClient(),
@@ -174,7 +177,7 @@ func main() {
 		Scheme:    mgr.GetScheme(),
 		Client:    mgr.GetClient(),
 		Cloud:     cld,
-		SCI:       gc,
+		SCI:       sciClient,
 		NewObject: func() controller.BuildableObject { return &apiv1.Notebook{} },
 		Kind:      "Notebook",
 	}).SetupWithManager(mgr); err != nil {
@@ -185,6 +188,7 @@ func main() {
 		Client: mgr.GetClient(),
 		Scheme: mgr.GetScheme(),
 		Cloud:  cld,
+		SCI:    sciClient,
 		ParamsReconciler: &controller.ParamsReconciler{
 			Scheme: mgr.GetScheme(),
 			Client: mgr.GetClient(),
@@ -197,7 +201,7 @@ func main() {
 		Scheme:    mgr.GetScheme(),
 		Client:    mgr.GetClient(),
 		Cloud:     cld,
-		SCI:       gc,
+		SCI:       sciClient,
 		NewObject: func() controller.BuildableObject { return &apiv1.Dataset{} },
 		Kind:      "Dataset",
 	}).SetupWithManager(mgr); err != nil {
--- a/internal/cloud/common.go
+++ b/internal/cloud/common.go
@@ -6,12 +6,15 @@ import (
 	"io"
 	"path/filepath"
 	"strings"
+
+	corev1 "k8s.io/api/core/v1"
 )
 
 type Common struct {
 	ClusterName       string     `env:"CLUSTER_NAME" validate:"required"`
 	ArtifactBucketURL *BucketURL `env:"ARTIFACT_BUCKET_URL,noinit" validate:"required"`
 	RegistryURL       string     `env:"REGISTRY_URL" validate:"required"`
+	Principal         string     `env:"PRINCIPAL"`
 }
 
 func (c *Common) ObjectBuiltImageURL(obj BuildableObject) string {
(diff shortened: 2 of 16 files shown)


cmd/controllermanager/main.go
```diff
@@ -174,7 +177,7 @@ func main() {
 		Scheme:    mgr.GetScheme(),
 		Client:    mgr.GetClient(),
 		Cloud:     cld,
-		SCI:       gc,
+		SCI:       sciClient,
```
brandonjbjelland, line 180: 👍

internal/cloud/gcp.go
```diff
@@ -116,11 +116,12 @@ func (gcp *GCP) MountBucket(podMetadata *metav1.ObjectMeta, podSpec *corev1.PodS
...
-	if sa.Annotations == nil {
-		sa.Annotations = map[string]string{}
-	}
-	sa.Annotations["iam.gke.io/gcp-service-account"] = fmt.Sprintf("substratus-%s@%s.iam.gserviceaccount.com", sa.Name, gcp.ProjectID)
+func (gcp *GCP) GetPrincipal(sa *corev1.ServiceAccount) string {
+	return fmt.Sprintf("substratus@%s.iam.gserviceaccount.com", gcp.ProjectID)
+}
+
+func (gcp *GCP) AssociatePrincipal(sa *corev1.ServiceAccount) {
+	sa.Annotations["iam.gke.io/gcp-service-account"] = gcp.GetPrincipal(sa)
```
brandonjbjelland, line 124: Hmm I don't love that we'd have provider specifics for annotations here. Potentially fine for now but I wonder what other options we have.

brandonjbjelland (dismissed):
overall lgtm. I'll give another pass later when I have more cycles

samos123:
Let's make sure tests pass before merging

internal/cloud/gcp.go
samos123: I think the only way to solve that would through K8s OSS to standardize on how workload identity should work

internal/controller/service_accounts.go
```diff
@@ -16,14 +17,33 @@ const (
 	modelServerServiceAccountName      = "model-server"
 	notebookServiceAccountName         = "notebook"
 	dataLoaderServiceAccountName       = "data-loader"
+	identityBoundLabel                 = "substratusai/identity-bound"
 )
 
-func reconcileCloudServiceAccount(ctx context.Context, cloudConfig cloud.Cloud, c client.Client, sa *corev1.ServiceAccount) (result, error) {
+func reconcileCloudServiceAccount(ctx context.Context, cloudConfig cloud.Cloud, sciClient sci.ControllerClient, c client.Client, sa *corev1.ServiceAccount) (result, error) {
```
brandonjbjelland, line 23: Aligning with naming elsewhere:
```suggestion
func reconcileCloudPrincipal(ctx context.Context, cloudConfig cloud.Cloud, sciClient sci.ControllerClient, c client.Client, sa *corev1.ServiceAccount) (result, error) {
```

samos123: I think we should rename to reconcileServiceAccount

internal/cloud/common.go
```diff
@@ -6,12 +6,15 @@ import (
...
 	"strings"
+
+	corev1 "k8s.io/api/core/v1"
 )
 
 type Common struct {
 	ClusterName       string     `env:"CLUSTER_NAME" validate:"required"`
 	ArtifactBucketURL *BucketURL `env:"ARTIFACT_BUCKET_URL,noinit" validate:"required"`
 	RegistryURL       string     `env:"REGISTRY_URL" validate:"required"`
+	Principal         string     `env:"PRINCIPAL"`
```
nstogner, line 17: I think this should be required (`validate:"required"`)?

internal/cloud/gcp.go
```diff
@@ -116,11 +116,12 @@ func (gcp *GCP) MountBucket(podMetadata *metav1.ObjectMeta, podSpec *corev1.PodS
 	return fmt.Errorf("container not found: %s", req.Container)
 }
 
-func (gcp *GCP) AssociateServiceAccount(sa *corev1.ServiceAccount) {
-	if sa.Annotations == nil {
-		sa.Annotations = map[string]string{}
-	}
-	sa.Annotations["iam.gke.io/gcp-service-account"] = fmt.Sprintf("substratus-%s@%s.iam.gserviceaccount.com", sa.Name, gcp.ProjectID)
+func (gcp *GCP) GetPrincipal(sa *corev1.ServiceAccount) string {
+	return fmt.Sprintf("substratus@%s.iam.gserviceaccount.com", gcp.ProjectID)
```
nstogner, line 120: This will result in the principal set in `PRINCIPAL` env variable never getting used as this method will overshadow the one from `Common`. Prefer to either leave this method off of `GCP` and lean on the one from `Common`. If you want to default to this email determined from `gcp.ProjectID` it should be done in `AutoConfigure()`. See how `RegistryURL` is set for reference.

internal/controller/service_accounts.go
nstogner: We reconcile the service account here (create/update the KSA)... as a part of doing that we reconcile the cloud principle. I like `reconcileServiceAccount` as the func name.

internal/controller/service_accounts.go
```diff
@@ -16,14 +17,33 @@ const (
...
+		bindIdentityRequest := sci.BindIdentityRequest{
+			Principal:                principal,
+			KubernetesServiceAccount: sa.Name,
+			KubernetesNamespace:      sa.Namespace,
+		}
+		if _, err := sciClient.BindIdentity(ctx, &bindIdentityRequest); err != nil {
+			return result{}, fmt.Errorf("failed bind identity principal %s to K8s SA %s/%s: %w",
+				principal, sa.Namespace, sa.Name, err)
+		}
+		sa.Annotations["identity-bound"] = principal
```
nstogner, line 44: This is ok, but it would be cleaner to re-use the cloud-specific annotation for the signal that this has been bound. How this could be done without passing around the cloud-specific annotation key? Perhaps `cloudConfig.GetPrinciple(sa)` could return two values `(string, bool)` with the 2nd being: whether this principal was already set on the given ServiceAccount object.

nstogner (changes requested):
Looks like the right solution. Requested a few changes.

internal/cloud/gcp.go
brandonjbjelland: Derp, this is a part of the Cloud interface - we're already doing exactly what I was suggesting here! Never mind me!

internal/controller/service_accounts.go
brandonjbjelland: Ahh yeah `reconcileServiceAccount` works best here

internal/cloud/common.go
samos123: the reason I made this optional is because that would require us to always set it, even when you use something like kind or when you run the tests. I had it as `validate:"required"` at first but after running tests, I thought it would be better to make it optional.

internal/controller/service_accounts.go
samos123: done, renamed to `reconcileServiceAccount`

internal/cloud/gcp.go
samos123: Fixed by using AutoConfigure. Thanks for catching this!

internal/controller/service_accounts.go
samos123: I think that would complicate it and make it less clean. It's  clearer what's happening here vs passing around  a label. 
Let me know if you feel strongly about it.

I did have a bug here because I wasn't using the `const` when setting the annotation.

internal/sci/fake_sci_client.go
brandonjbjelland, line 1: nice workaround!

brandonjbjelland (approved):
🍗

internal/cloud/common.go
samos123: After refactoring the PR, I think this could be changed back to required. Do you think that's the right move here? Even though Kind won't need it? There might be other environments where workload identity isn't available.

internal/sci/fake_sci_client.go
samos123: I'm curious if this might also remove the flakiness of tests since I noticed the Builder Controller required a CSI client but it was never passed.

brandonjbjelland: huh! How was it even compiling?

samos123: the value would be nill but it would compile

samos123: the reason I knew what to do was because I saw panics due to trying to call a method on a nil value

internal/cloud/common.go
nstogner: I see the benefit of requiring this (production environments need it) outweighing the negatives… dev envs needing to pass a stub value… alternative would be to have some custom validation logic, but prob not worth the effort there

internal/cloud/gcp_test.go
brandonjbjelland, line 1: Nice work on these!

brandonjbjelland (approved):
Yep 👍

internal/controller/service_accounts.go
samos123: I implemented your suggestion and agree that it's cleaner now. I misunderstood your comment originally. Great suggestion!

PR #190, opened 2023-08-12
Support for local kind cluster

### LINKED ISSUE #152: Local kind cluster installation
I would like to install the project in a kind cluster for local testing.

nstogner:
Add local kind support.

[See Diagram](https://raw.githubusercontent.com/substratusai/substratus/5c813780ccc3d139dd12586869dd85e3e9898fab/docs/diagrams/arch-kind.excalidraw.png)

Fixes #152


internal/sci/kind/server.go
```diff
@@ -0,0 +1,95 @@
...
+func (s *Server) saveUpload(r io.Reader, urlPath string) error {
+	path := filepath.Join(s.BucketDir, urlPath)
+	f, err := os.Create(path)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+	if _, err := io.Copy(f, r); err != nil {
+		return err
+	}
```
brandonjbjelland, line 56: Reaching a bit here but is it better to `os.Stat(dst)` the destination before copying so as to not overwrite?

nstogner: I think overwriting the stored tarball is what we do in the real buckets today

brandonjbjelland: > overwriting the stored tarball is what we do in the real buckets today

Huh - I bet we have a not-so-subtle bug on GCP where one user could easily overwrite another's before the builder can start a build against it.

nstogner: Are you thinking two users could both be running `kubectl notebook -d .` on the same Notebook at the same time?

nstogner: If so, we are not currently accounting for that. Notebooks are intended to be used by 1 user at a time at the moment.

internal/sci/kind/server.go
```diff
@@ -0,0 +1,110 @@
...
+	"log"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+
+	sci "github.com/substratusai/substratus/internal/sci"
+)
+
+var _ sci.ControllerServer = &Server{}
```
samos123, line 18: what is the reason for doing this? Trying to understand why this is needed

nstogner: Not needed, just creates a compile-time error if the `Server{}` does not satisfy the intended `sci.ControllerServer` interface.

config/install-kind/namespace.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
```
samos123, line 1: I think this should be included by default in all installation not just kind?

config/sci-kind/deployment.yaml
samos123, line 1: rename to `deployment_patch.yaml` to follow same pattern as with `manager_patch.yaml`

config/sci-kind/service.yaml
samos123, line 1: rename to `service_patch.yaml`

samos123 (changes requested):

cmd/sci-kind/main.go
```diff
@@ -0,0 +1,56 @@
...
+	var cfg struct {
+		port          int
+		signedURLPort int
+	}
+	flag.IntVar(&cfg.port, "port", 10080, "port number to listen on")
+	flag.IntVar(&cfg.signedURLPort, "signed-url-port", 8080, "port to listen for signed url traffic")
+	flag.Parse()
+
+	s := &scikind.Server{
+		SignedURLAddress: "http://localhost:30080",
```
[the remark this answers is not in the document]
nstogner: Agreed, I have already forgotten myself. Will document better

Makefile
```diff
@@ -142,6 +142,32 @@ dev-down-gcp: build-installer
...
+# TODO(nstogner): Running outside of cluster is tricky to support b/c of how substratus
+# Pods need to mount the same directories as the SCI.
+# 
+# .PHONY: dev-run-kind
+# dev-run-kind:
+# 	...
+#
+
+.PHONY: dev-skaffold-kind
+dev-skaffold-kind:
```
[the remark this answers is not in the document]
nstogner: Thought was: we might want a `dev-run-{cloud}` and `dev-skaffold-{cloud}` for each cloud. `dev-run-*` currently runs locally... There are probably some better names... Let me know your thoughts.

Makefile
```diff
@@ -142,6 +142,32 @@ dev-down-gcp: build-installer
...
+	# NOTE: Disabled cache artifacts can help if you are having issues
+	# with timeouts from docker hub. It coulld be quicker to just rebuild
+	# and rely on standard docker layer caching.
+    # --cache-artifacts=false
+	skaffold dev -f skaffold.kind.yaml --cache-artifacts=false
```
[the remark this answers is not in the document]
nstogner: Good callout

Makefile
```diff
@@ -257,7 +283,8 @@ installation-scripts:
 installation-manifests: manifests kustomize
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	cd config/gcpmanager && $(KUSTOMIZE) edit set image gcp-manager=${IMG_GCPMANAGER}
-	$(KUSTOMIZE) build config/default > install/kubernetes/system.yaml
+	$(KUSTOMIZE) build config/install-gcp > install/kubernetes/system.yaml
```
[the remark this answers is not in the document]
nstogner: Lets merge your PR 1st

internal/cloud/kind.go
```diff
@@ -0,0 +1,76 @@
...
+func (k *Kind) Name() string { return KindName }
+
+func (k *Kind) AutoConfigure(ctx context.Context) error {
+	// This environment variable comes from the registry Service in the same namespace.
+	// See: https://kubernetes.io/docs/concepts/services-networking/service/#environment-variables
+	addr := os.Getenv("REGISTRY_PORT_5000_TCP_ADDR")
+	if addr == "" {
+		return fmt.Errorf("REGISTRY_PORT_5000_TCP_ADDR not set")
+	}
+	k.RegistryURL = addr + ":5000"
```
samos123, line 28: Add auto configure for ArtifactURL instead of setting it in configmap

config/install-kind/namespace.yaml
samos123: I have done this in my PR/branch for now as well: https://github.com/substratusai/substratus/pull/185/commits/01d2ec10fd41dbf126348534888411eb0f43c1b6

Approach I took is add namespace.yaml under config/manager. I thought it wasn't worth creating a separate directory.

Makefile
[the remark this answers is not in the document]
samos123: in that case I like your naming, I like the convention of `dev-skaffold-{CLOUD}`. I guess the issue for kind is there is no way to do `dev-run-kind`? Maybe `dev-run-kind` just points to `dev-skaffold-kind`.

skaffold.kind.yaml
```diff
@@ -0,0 +1,16 @@
...
+build:
+  artifacts:
+    - image: docker.io/substratusai/controller-manager
+      docker:
+        dockerfile: Dockerfile
+    - image: docker.io/substratusai/sci-kind
+      docker:
+        dockerfile: Dockerfile.sci-kind
+  local:
+    push: false
```
samos123, line 16: how does this work, does kind automatically use the images built on the host? This shouldn't be working unless you do this: https://kind.sigs.k8s.io/docs/user/quick-start/#loading-an-image-into-your-cluster

samos123: found out why it works: https://skaffold.dev/docs/environment/local-cluster/

Makefile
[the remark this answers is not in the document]
samos123: I am now in favor of merging my PR first. I tested it very thoroughly and feel it's ready. Let's discuss more offline.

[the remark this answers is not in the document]
brandonjbjelland: Yeah it's become necessary at this point. I don't think we'll end up in this state in the future (developing multiple SCIs at once) but I think we can avoid this if we worked in smaller chunks as PRs (SCI, SCI implementation, deployment bits, and lastly the controller integration - in that order)

config/install-kind/manager_patch.yaml
```diff
@@ -0,0 +1,58 @@
+# This patch inject a sidecar container which is a HTTP proxy for the
+# controller manager, it performs RBAC authorization against the Kubernetes API using SubjectAccessReviews.
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: controller-manager
+  namespace: system
```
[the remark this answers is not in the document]
brandonjbjelland: Yeah holding off until your PR is merged.

samos123 (approved):

PR #197, opened 2023-08-17
Makefile improvements for GCP

samos123:
* The Makefile was slow because it often calls gcloud. This reduces the amount of gcloud calls required.
* Utilize the built-in skaffold that gets downloaded as part of the Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,6 @@ ENVTEST_K8S_VERSION = 1.26.1
 PLATFORM=$(shell uname -s | tr '[:upper:]' '[:lower:]')
 ARCH=$(shell uname -m | sed 's/x86_64/amd64/')
 
-PROJECT_ID := $(shell gcloud config get-value project)
 UNAME_S := $(shell uname -s)
 UNAME_M := $(shell uname -m)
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -158,10 +159,10 @@ dev-down-kind:
 	cd install/scripts && ./kind-down.sh
 
 .PHONY: dev-skaffold-gcp
-dev-skaffold-gcp: export PROJECT_ID=$(shell gcloud config get project)
+dev-skaffold-gcp: PROJECT_ID ?=$(shell gcloud config get project)
```
nstogner, line 162: Is this needed here? PROJECT_ID is set at top level now

samos123: it's not set at the top level and it shouldn't be because then it gets executed too many times and would fail if you're not on GCP?

Makefile
```diff
@@ -11,7 +11,6 @@ ENVTEST_K8S_VERSION = 1.26.1
 PLATFORM=$(shell uname -s | tr '[:upper:]' '[:lower:]')
 ARCH=$(shell uname -m | sed 's/x86_64/amd64/')
 
-PROJECT_ID := $(shell gcloud config get-value project)
```
samos123, line 14: @nstogner see it's no longer set at the top level. that's part of the improvement of this PR

brandonjbjelland (approved):
Essentially this change moves gcloud commands to just the make targets that use gcloud. Yeah that seems reasonable if you've seen slowdown and this reduces the calls. 👍

Makefile
nstogner: Mis-read, looks good

nstogner (approved):

PR #204, opened 2023-08-21
remove terraform from GCP install

samos123:
* updates gcp to use new directory structure
* creates gke cluster and nodepools during up.sh


Makefile
```diff
@@ -280,15 +269,15 @@ uninstall-crds: manifests kustomize ## Uninstall CRDs from the K8s cluster speci
...
+	perl -pi -e "s/version=.*/version=$(VERSION)/g" install/kubectl-plugins.sh
 
 .PHONY: installation-manifests
 installation-manifests: manifests kustomize
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	cd config/sci-kind && $(KUSTOMIZE) edit set image sci=${IMG_SCI_KIND}
 	$(KUSTOMIZE) build config/install-kind > install/kind/manifests/system.yaml
 	cd config/sci-gcp && $(KUSTOMIZE) edit set image sci=${IMG_SCI_GCP}
-	$(KUSTOMIZE) build config/install-gcp > install/kubernetes/gcp/system.yaml
+	$(KUSTOMIZE) build config/install-gcp > install/gcp/substratus.yaml
```
nstogner, line 280: `install/gcp/manifests.yaml`

install/gcp/up.sh
```diff
@@ -0,0 +1,115 @@
...
+REGION=${REGION:-us-central1}
+ZONE=${ZONE:=${REGION}-a}
+INSTALL_OPERATOR=no # set to yes if you want to install operator
+
+# Enable required services.
+gcloud services enable container.googleapis.com
+gcloud services enable artifactregistry.googleapis.com
+
+export CLUSTER_NAME=substratus
+gcloud container clusters create ${CLUSTER_NAME} --location ${REGION} \
```
nstogner, line 16: Looks like in some places we create, in some places we check then create if not exists... Perhaps we standardize on one or the other. Simplicity favors always creating, practically speaking though, it might be simpler for the user to be able to re-run the script on failure (idempotency). I am open to either.

install/gcp/up.sh
```diff
@@ -0,0 +1,115 @@
+#!/bin/bash
+
+set -u
+set -x
```
nstogner, line 4: Best to add `set -e` to exit on error?

nstogner (commented):
Really like the direction, I think you are missing `install/gcp/manifests.yaml`

install/gcp/up.sh
samos123: I removed it on purpose so I don't have to add a ton of if checks for gke cluster and nodepool creation. I can add it back and add if statements.

install/gcp/up.sh
samos123: I've added if statements everywhere

install/gcp/up.sh
samos123: I've added it back

nstogner (approved):

PR #205, opened 2023-08-21
Add system tests

nstogner:


samos123:
I would prefer to keep running this on every pull request push. What's the reason for not doing that?

PR #232, opened 2023-09-06
Standalone CLI: sub

nstogner:
Add a standalone CLI: `sub`

Replaces the current kubectl plugins.


Makefile
```diff
@@ -188,7 +188,7 @@ dev-down-aws: build-installer
 .PHONY: dev-run-gcp
 # Controller manager configuration #
 dev-run-gcp: export CLOUD=gcp
-dev-run-gcp: PROJECT_ID ?= $(shell gcloud config get project)
```
samos123, line 191: I purposely used `?=` so it doesn't get rerun every time. Any reason why you had to change it back to `export`?

api/v1/model_types.go
```diff
@@ -20,12 +20,12 @@ type ModelSpec struct {
 	// Resources are the compute resources required by the container.
 	Resources *Resources `json:"resources,omitempty"`
 
-	// BaseModel should be set in order to mount another model to be
+	// Model should be set in order to mount another model to be
 	// used for transfer learning.
-	BaseModel *ObjectRef `json:"baseModel,omitempty"`
+	Model *ObjectRef `json:"model,omitempty"`
```
samos123, line 25: I think this rename makes sense.

However, I am not a big fan of renames like these since those result in a lot of downstream changes to the examples and docs as well. The value we get out of these is medium but the amount of effort to validate the change doesn't break things is significant.

api/v1/model_types.go
```diff
@@ -20,12 +20,12 @@ type ModelSpec struct {
...
 
 	// Dataset to mount for training.
-	TrainingDataset *ObjectRef `json:"trainingDataset,omitempty"`
+	Dataset *ObjectRef `json:"dataset,omitempty"`
```
samos123, line 28: Same comment. I think this is a good rename and support it just calling out that we shouldn't take these lightly.

docs/cli.md
```diff
@@ -0,0 +1,104 @@
...
+## Notebook
+
+```bash
+strat nb (notebook)
+```
+
+## List
+
+```bash
+strat ls (list)
```
samos123, line 33: +1 to `strat`, -1 to `ai`

docs/cli.md
```diff
@@ -0,0 +1,104 @@
...
+* Remote build
+* Wait for Ready
+
+```bash
+strat run .
+```
+
+## View
+
+* Grab `run.html` (converted notebook) and serve on localhost.
```
samos123, line 76: can this be run only once ready or could we get results like these while the model is being ran? if so does it keep on converting the notebook to html every x seconds?

I will probably answer this myself based on this PR review

docs/container-contract.md
```diff
@@ -26,11 +26,10 @@ Note: This requirement is satisfied by default when using Substratus base images
 
 ```
 /content/         # Working directory.
-  data/           # Location where Datasets will be mounted (reading and loading).
+  data/           # Location where a previously stored Datasets is mounted.
+  model/          # Location where a previously stored Model is mounted.
```
samos123, line 30: where is the current model mounted?

Let's see I'm creating a fine tuned model where would that be mounted?

docs/container-contract.md
```diff
@@ -26,11 +26,10 @@ Note: This requirement is satisfied by default when using Substratus base images
...
+  output/         # Location to store output of a run.
   src/            # Source code (*.py, *.ipynb) for loading, training, etc.
-  logs/           # Output of building/training jobs for debugging purposes.
-  model/          # Location to store the resulting model from loading or training.
-  saved-model/    # Location where a previously saved model will be mounted.
```
samos123, line 33: What's the reason for no longer using /content/saved-model?

docs/design.md
```diff
@@ -166,37 +163,32 @@ runJob()
...
 ##### Model (training)
 
 ```
 /content/params.json        # Populated from .spec.params (also available as env vars).
 
-/content/data/              # Mounted RO: from .spec.trainingDataset
+/content/data/              # Mounted RO: from .spec.dataset
 
-/content/saved-model/       # Mounted RO: from .spec.baseModel
+/content/model/             # Mounted RO: from .spec.model
```
samos123, line 189: so which directory should be used for writing the new model files?

install/gcp/up.sh
```diff
@@ -15,89 +15,76 @@ gcloud services enable artifactregistry.googleapis.com
...
 if ! gcloud container clusters describe ${CLUSTER_NAME} --location ${REGION} -q >/dev/null; then
-gcloud container clusters create ${CLUSTER_NAME} --location ${REGION} \
-  --machine-type n2d-standard-8 --num-nodes 1 --min-nodes 1 --max-nodes 5 \
-  --node-locations ${ZONE} --workload-pool ${PROJECT_ID}.svc.id.goog \
-  --enable-image-streaming --enable-shielded-nodes --shielded-secure-boot \
-  --shielded-integrity-monitoring --enable-autoprovisioning \
-  --max-cpu 960 --max-memory 9600 --ephemeral-storage-local-ssd=count=2 \
-  --autoprovisioning-scopes=logging.write,monitoring,devstorage.read_only,compute \
-  --addons GcsFuseCsiDriver
+	gcloud container clusters create ${CLUSTER_NAME} --location ${REGION} \
```
samos123, line 18: this messes up the docs rendering so would prefer original formatting instead of so many spaces in front

samos123:
Overall the CLI looks promising! Can't wait to try it out for my dev flow.

I would prefer to keep this PR only to what's needed to make the Standalone CLI work. Any refactors should be done as separate PRs and discussed separately. Things like reformatting the gcp install scripts or renaming `baseModel` to `model` are big changes that don't seem to be required for the standalone CLI.

docs/cli.md
samos123: I think `sub` has my preference right now

docs/container-contract.md
```diff
@@ -26,11 +26,10 @@ Note: This requirement is satisfied by default when using Substratus base images
 
 ```
 /content/         # Working directory.
-  data/           # Location where Datasets will be mounted (reading and loading).
+  data/           # Location where a previously stored Datasets is mounted.
+  model/          # Location where a previously stored Model is mounted.
+  output/         # Location to store output of a run.
```
samos123, line 31: change to `artifacts

docs/design.md
```diff
@@ -127,7 +125,6 @@ The example Model's backing storage would end up being:
 
 ```sh
 gs://abc123-substratus-artifacts/f94a0d128bcbd9c1b824e9e5572baf86/model/
```
samos123, line 127: I think there is no longer a good reason to store model in a subdirectory

install/gcp/up.sh
samos123, line 1: please revert all changes in this file unless these changes were needed to fix something

install/gcp/up.sh
```diff
@@ -15,89 +15,76 @@ gcloud services enable artifactregistry.googleapis.com
...
+		--machine-type n2d-standard-8 --num-nodes 1 --min-nodes 1 --max-nodes 5 \
+		--node-locations ${ZONE} --workload-pool ${PROJECT_ID}.svc.id.goog \
+		--enable-image-streaming --enable-shielded-nodes --shielded-secure-boot \
+		--shielded-integrity-monitoring --enable-autoprovisioning \
+		--max-cpu 960 --max-memory 9600 --ephemeral-storage-local-ssd=count=2 \
+		--autoprovisioning-scopes=logging.write,monitoring,devstorage.read_only,compute \
+		--addons GcsFuseCsiDriver
 fi
 
-# Configure a maintenance exclusion to prevent automatic upgrades for 160 days
```
samos123, line 28: I've already fixed this in main

samos123:
I noticed the following log:
```
2023/09/09 21:34:03 Pod logs for: %v: %v test-gguf-model-notebook-bld-n85wk time="2023-09-10T04:34:03Z" level=info msg="Retrieving image nvidia/cuda:12.2.0-devel-ubuntu22.04 from registry index.docker.io"
```

samos123:
One thing that was rather annoying was being unable to edit params.json from within the notebook. It's mounted as read-only. I noticed I had a typo in my param.json and my only way to solve it was to delete the notebook and modify the substratus resource.

samos123:
I can't seem to get out once I hit an error, Ctrl + C and Q are both not working:
image

samos123:
file syncing currently only happens for files in `src` directory. I placed my notebook in the main directory. I think the file sync should happen for any  .ipynb file that's also in the local directory.

Makefile
samos123: seems it causes the project ID to be not set correct so that's why you changed it. Lets use what you got for now and I can figure out another way to only have to run `gcloud config get project` once.

docs/cli.md
```diff
@@ -0,0 +1,316 @@
...
+sub cl (classify) # Image recognition
+```
+
+https://github.com/charmbracelet/bubbletea/tree/master/examples#chat
+
+
+## Substratus.yaml
+
+### Option 1: Workspace file
+
```
samos123, line 164: workspace file seems complex, having multiple files feels cleaner

docs/cli.md
```diff
@@ -0,0 +1,316 @@
...
+**Pros:**
+
+* No new objects
+* Options for more than 1 objects
+* Ability to express remote base-model/dataset dependencies
+
+**Cons:**
+
+* Could get messy
```
samos123, line 278: and complex if you never used kustomize

docs/cli.md
```diff
@@ -0,0 +1,316 @@
...
+    name: facebook-opt-125m
+dataset:
+server:
+  # Generated if not specified
+  # Only valid if model is specified.
+notebook:
+  # Generated if not specified
+```
+
+### Option 2: Multi-doc
```
samos123, line 184: I think this or simply creating multiple files within same dir is better

containertools/cmd/nbwatch/main.go
```diff
@@ -33,7 +34,31 @@ func run() error {
 	}
 	defer w.Close()
 
-	w.Add("/content/src")
+	const contentDir = "/content"
+
+	// NOTE: Watch is non-recursive.
```
samos123, line 39: so it would watch `/content/src` but not `/content/src/my-dir/sam.ipynb`?

internal/controller/dataset_controller.go
```diff
@@ -176,10 +194,9 @@ func (r *DatasetReconciler) loadJob(ctx context.Context, dataset *apiv1.Dataset)
 	}
 
 	if err := r.Cloud.MountBucket(&job.Spec.Template.ObjectMeta, &job.Spec.Template.Spec, dataset, cloud.MountBucketConfig{
-		Name: "dataset",
+		Name: "artifacts",
 		Mounts: []cloud.BucketMount{
-			{BucketSubdir: "data", ContentSubdir: "data"},
-			{BucketSubdir: "logs", ContentSubdir: "logs"},
+			{BucketSubdir: "artifacts", ContentSubdir: "artifacts"},
```
samos123, line 199: Do we still need a `BucketSubdir` now that it is all artifacts together? I think probably cleaner to remove it

containertools/cmd/nbwatch/main.go
nstogner: Yep

internal/controller/dataset_controller.go
nstogner: I dont think we need it anymore

docs/cli.md
nstogner: agreed

docs/cli.md
nstogner: agreed

install/gcp/up.sh
nstogner: rebased

install/gcp/up.sh
nstogner: will do before merge

docs/design.md
nstogner: agreed, I think this doc is wrong now, will fix

docs/cli.md
nstogner: sub it is!

internal/tui/manifests.go
```diff
@@ -0,0 +1,195 @@
...
+		}
+
+		switch t := obj.(type) {
+		case *apiv1.Model, *apiv1.Dataset, *apiv1.Server, *apiv1.Notebook:
+			kind := t.GetObjectKind().GroupVersionKind().Kind
+			if m[kind] == nil {
+				m[kind] = make([]client.Object, 0)
+			}
+			m[kind] = append(m[kind], obj)
+			// case *apiv1.Model:
```
samos123, line 183: I would prefer to have these comments removed unless you feel there is a need to use this in the future?

internal/cli/push.go
```diff
@@ -0,0 +1,89 @@
...
+		}).New())
+		if _, err := tui.P.Run(); err != nil {
+			return err
+		}
+
+		return nil
+	}
+
+	cmd := &cobra.Command{
+		Use:   "push [dir]",
```
samos123, line 60: I'm fine with `push` as well but feel `run` is more descriptive of what the end-result is here. The pushing is just an intermediary step to run it.

nstogner: Updated to `run`

internal/controller/dataset_controller.go
nstogner: Will leave for later refactor

install/gcp/up.sh
nstogner: Editor side-effect, reverted

nstogner:
Tracking the param.json issue here: https://github.com/substratusai/substratus/issues/237

samos123 (approved):

PR #240, opened 2023-10-02
Fix system tests and run on PRs

samos123:
* Fixes the system tests which broke after CLI refactor
* Runs the system tests when PRs are triggered by using Skaffold to build and run the current revision


samos123:
Merging after verbal review of Nick

PR #248, opened 2023-10-08
fix #247 allow setting eph disk request

### LINKED ISSUE #247: Setting resources.disk doesn't seem to work
```
apiVersion: substratus.ai/v1
kind: Notebook
metadata:
  name: k8s-instruct-dataset
spec:
  suspend: true
  build:
    git:
      url: https://github.com/substratusai/images
      path: base
  resources:
    disk: 300
```

I was expecting a 300GB ephemeral disk request but doesn't seem like that happened

samos123:


samos123:
tested it by building image and running in GKE,  I verified that with this PR it does indeed set ephemeral disk correctly

internal/controller/notebook_controller.go
```diff
@@ -431,14 +431,16 @@ func (r *NotebookReconciler) notebookPod(notebook *apiv1.Notebook, model *apiv1.
...
-	// Mounts specific to this Notebook.
-	if err := r.Cloud.MountBucket(&pod.ObjectMeta, &pod.Spec, notebook, cloud.MountBucketConfig{
-		Name:      "artifacts",
-		Mounts:    []cloud.BucketMount{{BucketSubdir: "artifacts", ContentSubdir: "artifacts"}},
-		Container: containerName,
-		ReadOnly:  false,
-	}); err != nil {
-		return nil, fmt.Errorf("mounting notebook: %w", err)
+	// Mounts specific to this Notebook
+	if notebook.Spec.Dataset != nil || notebook.Spec.Model != nil {
```
nstogner, line 435: Why not mount artifacts if dataset and model are not defined?

samos123: so silly GCS fuse didn't get in the way lol. My thinking was if there is nothing to mount then we don't need to mount anything. I've reverted this back

internal/resources/resources.go
```diff
@@ -33,6 +33,7 @@ func Apply(podMetadata *metav1.ObjectMeta, podSpec *corev1.PodSpec, containerNam
 
 	resources.Requests[corev1.ResourceCPU] = *resource.NewQuantity(res.CPU, resource.DecimalSI)
 	resources.Requests[corev1.ResourceMemory] = *resource.NewQuantity(res.Memory*gigabyte, resource.BinarySI)
+	resources.Requests[corev1.ResourceEphemeralStorage] = *resource.NewQuantity(res.Disk*gigabyte, resource.BinarySI)
```
nstogner, line 36: Should we also add a default disk resource request when .resources is not defined? See `if res == nil` conditional above.

samos123: Yes great catch. I should probably add some tests

nstogner (approved):

PR #258, opened 2023-10-16
Add Runbooks proposal

nstogner:


samos123:
My initial thought on this that this relations seem a bit complex. I'm having a hard time understanding these concepts of Runbook and Run. I really like the approach of having Model, Dataset and Notebook. I think the only thing lacking is a way to run a Notebook resource as a Job and possibly introduce another CRD like an Evaluator.

PR #260, opened 2023-11-21
update README

samos123:


nstogner:
Perhaps add a banner about the current status: it will be refactored.

